PR #17, opened 2026-02-17
fix: add typescript as dependency to resolve knip peer dep in npx/dlx environments

olivermontes:
## Problem

When running `npx -y react-doctor@latest .` or `pnpm dlx react-doctor@latest .`, the following error occurs:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'typescript' imported from
.../knip/dist/plugins/astro/resolveFromAST.js
```

## Root Cause

`knip` declares `typescript >=5.0.4` as a **required peer dependency** for its Astro plugin. When `react-doctor` is executed via `npx` or `pnpm dlx`, it runs in an isolated cache directory that does not have access to the host project's `node_modules`. As a result, `typescript` cannot be resolved at module load time.

## Fix

Add `typescript` as an explicit dependency in `packages/react-doctor/package.json`, matching the range required by knip (`>=5.0.4 <7`). This ensures `typescript` is always installed alongside `react-doctor` in any environment.

## Testing

Before fix:
```bash
npx -y react-doctor@latest .  # ERR_MODULE_NOT_FOUND for typescript
```

After fix: react-doctor runs correctly since typescript is bundled as a direct dependency.

### DIFF
--- a/packages/react-doctor/package.json
+++ b/packages/react-doctor/package.json
@@ -54,7 +54,8 @@
     "ora": "^9.3.0",
     "oxlint": "^1.47.0",
     "picocolors": "^1.1.1",
-    "prompts": "^2.4.2"
+    "prompts": "^2.4.2",
+    "typescript": ">=5.0.4 <7"
   },
   "devDependencies": {
     "@types/prompts": "^2.4.9",


aidenybai:
lgtm

ameetmadan:
not working on version 0.0.17

➜  web git:(main) ✗ npx -y react-doctor@latest .
node:internal/modules/package_json_reader:268
  throw new ERR_MODULE_NOT_FOUND(packageName, fileURLToPath(base), null);
        ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'typescript' imported from .npm/_npx/81e833f6d16d6127/node_modules/knip/dist/PrincipalFactory.js
    at Object.getPackageJSONURL (node:internal/modules/package_json_reader:268:9)
    at packageResolve (node:internal/modules/esm/resolve:768:81)
    at moduleResolve (node:internal/modules/esm/resolve:858:18)
    at defaultResolve (node:internal/modules/esm/resolve:990:11)
    at #cachedDefaultResolve (node:internal/modules/esm/loader:768:20)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:745:38)
    at ModuleLoader.getModuleJobForImport (node:internal/modules/esm/loader:318:38)
    at #link (node:internal/modules/esm/module_job:208:49) {
  code: 'ERR_MODULE_NOT_FOUND'
}

Node.js v24.8.0

still getting the same issue

PR #28, opened 2026-02-18
fix(website): prevent hydration mismatch in terminal demo

SultokTheF:
## Summary

Fix hydration mismatch on the website terminal demo.

The issue was caused by reading `localStorage` during initial render, which could produce different first render output between SSR and client (empty command on server vs completed command on client).

## What changed

- `Terminal` now always starts from `INITIAL_STATE` on first render.
- Moved persisted state check (`didAnimationComplete`) into `useEffect` (client-only after mount).
- If animation was already completed, state is set to `COMPLETED_STATE` after mount.

## Why this fixes it

SSR and first client render are now deterministic and identical, so React hydration no longer sees text/content mismatch.

## Files

- `packages/website/src/components/terminal.tsx`

## Validation

- Reproduced mismatch before fix.
- Verified render flow after fix: no server/client initial state divergence.

## Additional changes

- Ignore `package-lock.json` in repo/website gitignore files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,7 @@ dist
 .DS_Store
 .agents
 .cursor
+
+
+# MARK: You have ignore this
+package-lock.json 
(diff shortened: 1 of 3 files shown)


aidenybai:
Reimplemented in main — thanks for the fix!

PR #30, opened 2026-02-18
feat: add no-inline-object-props performance rule

theakshaymore:
## Summary
Adds a new performance rule `no-inline-object-props` that detects inline object and array literals passed directly as JSX props.

## Problem
When objects or arrays are created inline as JSX props, a new reference is created on every render. This causes unnecessary re-renders of child components even when the actual data hasn't changed — a common and silent React performance issue.

## Example

❌ Bad — new reference created on every render
```jsx
<Component style={{ color: 'red' }} />
<Component data={[1, 2, 3]} />
<Component config={{ theme: 'dark', size: 'lg' }} />
```

✅ Good — stable reference, no unnecessary re-renders
```jsx
const style = { color: 'red' };
<Component style={style} />
```

## Changes
- Added rule visitor in `performance.ts`
- Registered rule in `index.ts`
- Added severity (`warn`) in `oxlint-config.ts`
- Added category mapping and help text in `run-oxlint.ts`
- Added test coverage in `run-oxlint.test.ts`

## Validation
- All 81 tests pass (`pnpm --filter react-doctor test`)
- Typecheck passes
- Build passes
- Rule emits 7 diagnostics on fixture with correct severity/category

### DIFF
--- a/packages/react-doctor/src/oxlint-config.ts
+++ b/packages/react-doctor/src/oxlint-config.ts
@@ -119,6 +119,7 @@ export const createOxlintConfig = ({
     "react-doctor/rerender-memo-with-default-value": "warn",
     "react-doctor/rendering-animate-svg-wrapper": "warn",
     "react-doctor/no-inline-prop-on-memo-component": "warn",
+    "react-doctor/no-inline-object-props": "warn",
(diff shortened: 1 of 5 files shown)


aidenybai:
Closing — too noisy, would flag most JSX

PR #33, opened 2026-02-18
Detect react compiler in expo project

pchalupa:
Adds detection of React Compiler in Expo projects, see: https://docs.expo.dev/guides/react-compiler/#enabling-react-compiler.


aidenybai:
ty

PR #36, opened 2026-02-18
Pass ignore.files patterns to scanning tools

Nefleex:
## Summary

Pass `ignore.files` config patterns down to `runOxlint` and `runKnip` so excluded files are skipped during scanning, not just filtered from results after the fact.

## Motivation

The `ignore.files` config option is documented and accepted, but it only post-filters diagnostics in `filterIgnoredDiagnostics`. The scanning tools (oxlint and knip) still process every file regardless. This means projects with large build artifact directories still get hundreds of false positive "unused file" warnings from generated output, even when those paths are configured to be ignored.

## Usage

```json
{
  "ignore": {
    "files": ["build/**", ".cache/**"]
  }
}
```

With this change, the patterns above are now forwarded to:
- **oxlint** via `--ignore-pattern` CLI flags
- **knip** via post-filter using the existing `compileGlobPattern` utility

### DIFF
--- a/packages/react-doctor/src/scan.ts
+++ b/packages/react-doctor/src/scan.ts
@@ -361,6 +361,7 @@ export const scan = async (directory: string, inputOptions: ScanOptions = {}): P
     includePaths: inputOptions.includePaths,
   };
 
+  const ignoreFilePatterns = userConfig?.ignore?.files ?? [];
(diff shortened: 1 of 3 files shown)


aidenybai:
Superseded — fixed internally in #108/#110

PR #58, opened 2026-02-19
remove space from line label so they are clickable

AlexanderHott:
The current formatting for file + lines isn't picked up by the vscode file regex, meaning you can't click it to navigate directly to the location.

This PR changes the formatting from 

```
path/to/file: 42, 44, 46
```

to

```
path/to/file:42
path/to/file:44
path/to/file:46
```


aidenybai:
Reimplemented in main — thanks!

PR #64, opened 2026-02-20
feat: add Expo and React Native framework detection and enable RN-specific rules

athul-binu:
## Summary

React Native and Expo projects are currently detected as `"unknown"` framework (displayed as just "React"), and the 8 existing `rn-*` lint rules in the plugin are **never enabled** in the oxlint config — they're effectively dead code.

This PR:

- **Adds `"expo"` and `"react-native"` as recognized `Framework` types** in `types.ts`
- **Detects `expo` and `react-native` packages** in `FRAMEWORK_PACKAGES` during project discovery, so the CLI correctly displays "Expo" or "React Native" instead of "React"
- **Conditionally enables all 8 `rn-*` rules** when framework is Expo or React Native (same pattern as `NEXTJS_RULES`):
  - `rn-no-raw-text` (error) — raw text outside `<Text>` crashes on RN
  - `rn-no-deprecated-modules` (error) — removed modules from react-native core
  - `rn-no-legacy-expo-packages` (warn) — deprecated Expo packages
  - `rn-no-dimensions-get` (warn) — non-reactive dimension API
  - `rn-no-inline-flatlist-renderitem` (warn) — perf issue with inline renderItem
  - `rn-no-legacy-shadow-styles` (warn) — legacy shadow props vs boxShadow
  - `rn-prefer-reanimated` (warn) — UI-thread animations over JS-thread
  - `rn-no-single-element-style-array` (warn) — unnecessary array allocation
- **Adds "React Native" category** for all `rn-*` rules in diagnostic output
- **Adds actionable help text** for each `rn-*` rule

### Before
```
✓ Detecting framework. Found React.     ← wrong
  (no RN-specific diagnostics)           ← rules exist but never fire
```

### After (tested on a real Expo project)
```
✓ Detecting framework. Found Expo.
✗ "SafeAreaView" was removed from react-native — use react-native-safe-area-context instead
    Import from the community package instead — deprecated modules were removed from the react-native core
⚠ Dimensions.get() does not update on screen rotation or resize — use useWindowDimensions()
    Use const { width, height } = useWindowDimensions() — it updates reactively on rotation and resize
```

## Test plan

- [x] `pnpm build` succeeds
- [x] All 125 existing tests pass (`pnpm test`)
- [x] No new files created — follows existing patterns exactly (mirrors `NEXTJS_RULES` approach)
- [x] Verified on a real Expo project: framework shows as "Expo" and RN rules fire correctly

### DIFF
--- a/packages/react-doctor/src/oxlint-config.ts
+++ b/packages/react-doctor/src/oxlint-config.ts
@@ -22,6 +22,17 @@ const NEXTJS_RULES: Record<string, string> = {
   "react-doctor/nextjs-no-side-effect-in-get-handler": "error",
 };
 
+const REACT_NATIVE_RULES: Record<string, string> = {
+  "react-doctor/rn-no-raw-text": "error",
+  "react-doctor/rn-no-deprecated-modules": "error",
+  "react-doctor/rn-no-legacy-expo-packages": "warn",
+  "react-doctor/rn-no-dimensions-get": "warn",
+  "react-doctor/rn-no-inline-flatlist-renderitem": "warn",
+  "react-doctor/rn-no-legacy-shadow-styles": "warn",
+  "react-doctor/rn-prefer-reanimated": "warn",
+  "react-doctor/rn-no-single-element-style-array": "warn",
+};
+
 const REACT_COMPILER_RULES: Record<string, string> = {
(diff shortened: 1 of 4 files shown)


aidenybai:
good catch, ty

PR #78, opened 2026-02-23
fix: strip ANSI escape codes from PR comment output

k22036:
## Problem

When react-doctor output is posted as a GitHub PR comment, ANSI color codes (e.g. `\x1b[32m`) appear as garbled text because GitHub Markdown does not render ANSI escape sequences.

## Solution

Pipe the raw output through `sed` to strip all ANSI escape sequences before writing to the file used for the PR comment. The terminal output in the Actions log remains unaffected (colors are preserved via the raw file).

## Changes

- `action.yml`: Save raw output to a temporary file, then produce a clean copy with ANSI codes stripped for use in the PR comment body.

・before
スクリーンショット 2026-02-23 14 02 29

・after
スクリーンショット 2026-02-23 14 03 33

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -54,7 +54,8 @@ runs:
         if [ -n "$INPUT_DIFF" ]; then FLAGS="$FLAGS --diff $INPUT_DIFF"; fi
 
         if [ -n "$INPUT_GITHUB_TOKEN" ]; then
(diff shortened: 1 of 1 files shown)


aidenybai:
Duplicate of #80

PR #107, opened 2026-03-22
feat(action): add offline input to skip remote API calls

WillHen:
When running React Doctor in GitHub Actions, there's no way to disable publishing diagnostics to the react.doctor API. This is a problem for orgs where `react.doctor` is blocked by network security policies, self-hosted runners without external internet access, or teams that simply want local-only scoring.

## Changes

- Add `offline` input to `action.yml` (default `"false"`) — passes the existing `--offline` CLI flag when set to `"true"`
- Wire `--offline` into both the main scan step and the score extraction step
- Add `offline` and `fail-on` (previously missing) to the README inputs table

## Usage

```yaml
- uses: millionco/react-doctor@main
  with:
    offline: "true"
    diff: main
```

When `offline` is enabled, scores are calculated locally instead of via the remote API. All scanning behavior remains identical — only the score API calls are skipped.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -23,6 +23,9 @@ inputs:
   fail-on:
     description: "Exit with error code on diagnostics: error, warning, none"
     default: "error"
+  offline:
+    description: "Skip sending diagnostics to the react.doctor API and calculate score locally"
+    default: "false"
   node-version:
     description: "Node.js version to use"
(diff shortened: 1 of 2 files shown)


aidenybai:
Reimplemented in main — thanks!

PR #118, opened 2026-04-06
fix: detect namespace hook calls (React.useEffect, React.useState)

Akshay090:
## Summary

Fixes #117

- `isHookCall` only checked `Identifier` callee nodes, silently missing `MemberExpression` calls like `React.useEffect(...)` and `React.useState(...)`
- Added `getCalleeName` helper that extracts function names from both `Identifier` and `MemberExpression` nodes
- Updated all setter detection sites (`countSetStateCalls`, `noDerivedStateEffect`, `rerenderFunctionalSetstate`, `renderingHydrationNoFlicker`) to use the same pattern

## Test plan

- [x] `pnpm build` — passes (3 entry points built successfully)
- [x] Unit tests for `getCalleeName` — 5/5 passing (Identifier, MemberExpression, null, unsupported types)
- [x] Unit tests for `isHookCall` — 7/7 passing (direct calls, namespace calls, string vs Set hookName, non-CallExpression rejection)
- [x] Unit tests for `countSetStateCalls` — 3/3 passing (direct setters, namespace setters, non-setter calls)
- [x] `npx react-doctor . --verbose` on namespace imports (`import * as React`) — `no-derived-state-effect`, `no-fetch-in-effect`, `no-cascading-set-state` all fire on `React.useEffect`/`React.useState` calls
- [x] `npx react-doctor . --verbose` on named imports (`import { useEffect }`) — same rules fire, existing behavior unchanged

### DIFF
--- a/packages/react-doctor/src/plugin/helpers.ts
+++ b/packages/react-doctor/src/plugin/helpers.ts
@@ -55,12 +55,11 @@ export const getCallbackStatements = (callback: EsTreeNode): EsTreeNode[] => {
 export const countSetStateCalls = (node: EsTreeNode): number => {
   let setStateCallCount = 0;
   walkAst(node, (child) => {
-    if (
-      child.type === "CallExpression" &&
-      child.callee?.type === "Identifier" &&
-      isSetterIdentifier(child.callee.name)
-    ) {
-      setStateCallCount++;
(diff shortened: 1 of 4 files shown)


aidenybai:
Superseded — fixed internally in #120

PR #139, opened 2026-04-29
fix: clarify passive listener warning

### LINKED ISSUE #138: Clarify error message to prevent breaking functionality when `preventDefault()` is needed
**Description:**
The lint rule for passive event listeners (`react-doctor/client-passive-event-listeners`) suggests adding `{ passive: true }` without providing a critical warning about its side effects. This can lead developers to unknowingly introduce bugs.

**Current Behavior**
When an event listener is flagged, the error message suggests adding the `{ passive: true }` option as a fix.

**Problem**

While adding `{ passive: true }` often improves performance, it causes the browser to ignore any calls to `event.preventDefault()` within the event handler. The current error message doesn't mention this risk. ([Docs](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#passive))

As a result, a developer might apply the suggested fix without realizing it will break functionality, leading to unexpected behavior and regressions 

**Suggested Improvement:**
The error message should be updated to explicitly warn developers about this behavior so they can apply the fix safely.

**Proposed message:**
> Add `{ passive: true }` as the third argument: `addEventListener('scroll', handler, { passive: true })`. **Only do this if the handler does not call `event.preventDefault()`.**

zerone0x:
## Summary
Fixes #138

Clarifies the passive event listener guidance so developers are warned not to apply `{ passive: true }` when the handler calls `event.preventDefault()`.

## Changes
- Update the CLI help text for `client-passive-event-listeners`.
- Update the plugin diagnostic message to include the same `preventDefault()` caveat.

## Test Plan
- `git diff --check`
- `npx -y -p node@22.18.0 -p pnpm@10.29.1 pnpm typecheck`
- `npx -y -p node@22.18.0 -p pnpm@10.29.1 pnpm lint` (passes with pre-existing warnings)
- `npx -y -p node@22.18.0 -p pnpm@10.29.1 pnpm format`
- `npx -y -p node@22.18.0 -p pnpm@10.29.1 pnpm test`

---

> [!NOTE]
> **Low Risk**
> Text-only changes to lint diagnostics/help output; no behavior or rule matching logic is altered.
> 
> **Overview**
> Clarifies the `client-passive-event-listeners` diagnostic/help text to warn that `{ passive: true }` should only be added when the handler does *not* call `event.preventDefault()`.
> 
> Refactors the rule’s reported message to reuse a shared constant so the updated wording is consistent across reports.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a17ff3cbd10ca3a4314cd45ec2d38fad022330d7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/react-doctor/src/plugin/rules/client.ts
+++ b/packages/react-doctor/src/plugin/rules/client.ts
@@ -2,6 +2,9 @@ import { PASSIVE_EVENT_NAMES } from "../constants.js";
 import { isMemberProperty } from "../helpers.js";
 import type { EsTreeNode, Rule, RuleContext } from "../types.js";
 
+const PASSIVE_EVENT_LISTENER_MESSAGE =
+  "Listener without { passive: true } — improves scrolling performance, but only use it when the handler does not call event.preventDefault()";
+
 export const clientPassiveEventListeners: Rule = {
   create: (context: RuleContext) => ({
     CallExpression(node: EsTreeNode) {
@@ -17,7 +20,7 @@ export const clientPassiveEventListeners: Rule = {
       if (!optionsArgument) {
         context.report({
           node,
-          message: `"${eventName}" listener without { passive: true } — blocks scrolling performance`,
+          message: `"${eventName}" ${PASSIVE_EVENT_LISTENER_MESSAGE}`,
         });
         return;
       }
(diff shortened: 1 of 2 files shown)


aidenybai:
Already shipped on main via bab66d0 (fix(client-passive-event-listeners): warn about preventDefault() incompatibility (#138)). Closing as already merged via direct commit — thanks!

PR #209, opened 2026-05-12
fix(nextjs): ignore response header mutations in GET handlers

### LINKED ISSUE #206: False positive: nextjs-no-side-effect-in-get-handler flags Response.headers.set() and similar Headers API calls
## Summary

## Reproduction

```ts
  return res;
}
```

## Expected behavior

The rule should NOT fire on:

## Workaround (current)

## Environment

- `react-doctor` v0.1.6
- Next.js 14 App Router
- Node 20

aqilaziz:
﻿## Summary
- ignore outbound response header mutations like `response.headers.set()` in GET route side-effect detection
- ignore request-scoped `Headers`/`Map` mutations created inside the handler
- add regression coverage for issue #206 using a temporary Next.js route fixture

Fixes #206.

## Testing
- `pnpm exec vp test run tests/regressions/nextjs-side-effects.test.ts`
- `pnpm exec tsc --noEmit` from `packages/react-doctor`
- `pnpm exec vp lint`
- `git diff --check`

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are limited to lint-side-effect heuristics and test fixtures, primarily reducing false positives without affecting runtime behavior.
> 
> **Overview**
> Refines `nextjs-no-side-effect-in-get-handler` detection by teaching `findSideEffect` to **ignore outbound response header shaping** (e.g. `response.headers.set/append/delete` and `headers().*` mutations) and **ignore mutations on request-scoped collections** (`new Headers/Map/Set` created within the handler).
> 
> Adds a Next.js route fixture and a regression test ensuring these patterns no longer produce diagnostics, and includes a patch changeset for `react-doctor`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bcde71350a56c8f3f0f6b23d6e287fd74eaeeaa0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/quiet-headers-scope.md
+++ b/.changeset/quiet-headers-scope.md
@@ -1,0 +1,5 @@
+---
+"react-doctor": patch
+---
+
+Avoid reporting response header mutations and request-scoped collections as GET handler side effects.
--- a/packages/oxlint-plugin-react-doctor/src/plugin/constants/library.ts
+++ b/packages/oxlint-plugin-react-doctor/src/plugin/constants/library.ts
@@ -28,4 +28,8 @@ export const MUTATION_METHOD_NAMES = new Set([
   "append",
 ]);
 
+export const HEADERS_API_MUTATION_METHOD_NAMES = new Set(["append", "delete", "set"]);
+
+export const REQUEST_SCOPED_MUTATION_CONSTRUCTOR_NAMES = new Set(["Headers", "Map", "Set"]);
+
 export const MUTATING_HTTP_METHODS = new Set(["POST", "PUT", "DELETE", "PATCH"]);
--- a/packages/oxlint-plugin-react-doctor/src/plugin/utils/find-side-effect.ts
+++ b/packages/oxlint-plugin-react-doctor/src/plugin/utils/find-side-effect.ts
@@ -1,4 +1,9 @@
-import { MUTATING_HTTP_METHODS, MUTATION_METHOD_NAMES } from "../constants/library.js";
+import {
+  HEADERS_API_MUTATION_METHOD_NAMES,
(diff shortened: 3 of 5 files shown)


aqilaziz:
Rebased/rebuilt this PR onto current `main` after the monorepo refactor, so GitHub now reports it as mergeable.

What changed in the updated branch:
- moved the fix to the new `oxlint-plugin-react-doctor` rule utility location
- kept the same behavior: ignore response header shaping (`response.headers.set/append/delete`) and request-scoped `Headers`/`Map`/`Set` mutations in GET handlers
- added a Next.js fixture route and regression assertion for the headers case

Verification:
- `pnpm -C packages/oxlint-plugin-react-doctor typecheck`
- built the internal packages on Windows using PowerShell env syntax for `NODE_ENV`
- direct `runOxlint` check against the new `app/headers/route.tsx` fixture returns no `nextjs-no-side-effect-in-get-handler` diagnostics

Note: `pnpm -C packages/react-doctor test -- tests/run-oxlint/nextjs.test.ts` currently still has unrelated baseline failures on current `main` in this checkout, including existing Next.js rule expectations. Vercel is also blocked by team authorization as before.

NisargIO:
Working on a related fix in #260 that takes a slightly broader angle — covers `.headers` chains, locally-scoped `Map/Set/Headers/URLSearchParams/FormData`, `.searchParams`, aliased `headers()`, cron routes, and depth-bounded handler resolution for `export const GET = withAuth(handler)` patterns. Thanks for kicking this off — your repro and test cases were super useful while validating the broader fix.

PR #233, opened 2026-05-14
fix(#206): skip response.headers.set() in nextjs-no-side-effect rule

william-garden:
## Summary

Issue #206: False positive - nextjs-no-side-effect-in-get-handler flags Response.headers.set()

## Problem

The `nextjs-no-side-effect-in-get-handler` rule was incorrectly flagging `response.headers.set()` calls inside GET route handlers as security-risk side effects. These calls only shape the outbound response (Cache-Control, X-Deprecated, CORS headers) and are idiomatic Next.js — they do not mutate server state.

## Solution

Added `isResponseHeadersCall()` function to detect `response.headers.set()`, `response.headers.append()`, and `response.headers.delete()` calls and skip them during side effect detection.

## Changes

- `packages/react-doctor/src/plugin/utils/is-response-headers-call.ts` (new)
- `packages/react-doctor/src/plugin/utils/find-side-effect.ts` - skip response.headers calls
- `packages/react-doctor/src/plugin/utils/index.ts` - export new function

---

> [!NOTE]
> **Low Risk**
> Low risk predicate change to a lint rule to reduce false positives; main risk is potentially missing a real side effect if code matches the new `response.headers.*` call shape unexpectedly.
> 
> **Overview**
> Updates side-effect detection used by `nextjs-no-side-effect-in-get-handler` to *ignore* `response.headers.set()`, `.append()`, and `.delete()` calls so response header shaping is no longer reported as a server-state mutation.
> 
> Adds a new `isResponseHeadersCall` AST predicate and exports it via the utils index, then short-circuits `findSideEffect` when this call pattern is encountered.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a5bacbb1b44f2cb26e96b3e29ecb135d1ccbf0e0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/react-doctor/src/plugin/utils/find-side-effect.ts
+++ b/packages/react-doctor/src/plugin/utils/find-side-effect.ts
@@ -3,6 +3,7 @@ import { isCookiesOrHeadersCall } from "./is-cookies-or-headers-call.js";
 import { isMutatingDbCall } from "./is-mutating-db-call.js";
 import { isMutatingFetchCall } from "./is-mutating-fetch-call.js";
 import { isMutatingMethodProperty } from "./is-mutating-method-property.js";
+import { isResponseHeadersCall } from "./is-response-headers-call.js";
 import { walkAst } from "./walk-ast.js";
 import { isNodeOfType } from "./is-node-of-type.js";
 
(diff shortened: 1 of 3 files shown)


NisargIO:
Working on a related fix in #260 that takes a slightly broader angle — covers `response.headers.set/append/delete` (your case) plus the wider class: any chain ending in `.headers`, locally-scoped `Map/Set/Headers/URLSearchParams/FormData` bindings, `.searchParams`, aliased `headers()`, cron routes, and depth-bounded handler resolution for `export const GET = withAuth(handler)` patterns. Thanks for the report — your repro and test cases were super useful while validating the broader fix.

PR #240, opened 2026-05-14
fix: support member expressions in containsAuthCheck

### LINKED ISSUE #239: False positives on auth gating server functions

```ts
  "auth",
  "getSession",
  "getServerSession",
  "getUser",
  "requireAuth",
  "checkAuth",
  "verifyAuth",
  "authenticate",
  "currentUser",
  "getAuth",
  "validateSession",
]);

-- 
```

This will not match e.g.
```ts
auth0.getSession()
```

cr4yfish:
# Support member expressions for server auth rule

Fixes this case:
```ts
await auth0.getSession() // Currently undetected

// Current workaround
const getSession = async () => await auth0.getSession()
await getSession() // Gets detected
```

I just added an else if to catch that - small diff.

Closes #239

---

> [!NOTE]
> **Low Risk**
> Low risk, localized change to AST pattern matching for auth-check detection; main risk is minor false positives/negatives in linting rather than runtime behavior.
> 
> **Overview**
> Updates `containsAuthCheck` in `server-auth-actions.ts` to treat member-expression calls like `auth0.getSession()` (and awaited variants) as valid auth checks, in addition to direct identifier calls such as `getSession()`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 48965d85ef20a461a8953184bf96d3ffb471e64a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/react-doctor/src/plugin/rules/server/server-auth-actions.ts
+++ b/packages/react-doctor/src/plugin/rules/server/server-auth-actions.ts
@@ -38,7 +38,7 @@ const containsAuthCheck = (statements: EsTreeNode[]): boolean => {
       // Indirect call e.g. auth0.getSession()
       else if (
         isNodeOfType(callee, "MemberExpression") &&
-        isNodeOfType(callee.property, "Identified") &&
+        isNodeOfType(callee.property, "Identifier") &&
         AUTH_FUNCTION_NAMES.has(callee.property.name)
        ) {
         foundAuthCall = true;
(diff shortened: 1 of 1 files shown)


NisargIO:
Working on a related fix in #261 that uses the existing `getCalleeName` utility instead of adding a parallel `else if` branch — it already resolves both Identifier and MemberExpression callees to a single name, so chained members (`supabase.auth.getSession()`), optional chaining (`auth0?.getSession()`), and the original Identifier forms all collapse into one check. Thanks for the report and repro — your test case is locked into the regression suite there.

PR #238, opened 2026-05-14
Fix/issue 216 react review audit

william-garden:
fix(#216): resolve React Review audit - 1 error + 8 warnings

- effect-needs-cleanup: add clearTimeout in animated-score.tsx
- nextjs-no-img-element: use next/image in badge-snippet, leaderboard, terminal
- nextjs-missing-metadata: add metadata export in page.tsx
- async-await-in-loop: add eslint-disable comments for intentional sequential animations
- no-inline-exhaustive-style: extract inline styles to constants in og/route.tsx

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes how `react-doctor` detects side effects in GET handlers (security linting), which could alter reported violations; website changes are mostly safe lint/cleanup fixes.
> 
> **Overview**
> **Refines `react-doctor` side-effect detection in GET handlers** by ignoring `response.headers.{set,append,delete}()` calls (new `isResponseHeadersCall` helper) so response-header writes don’t get flagged as server-state mutations.
> 
> **Cleans up Next.js website audit warnings** by switching remaining `` usages to `next/image`, adding missing `metadata` to the home page, adding a `clearTimeout` cleanup in `AnimatedScore`, annotating intentional `await`-in-loop animations in `Terminal`, and extracting OG image inline styles into constants.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2c3cbaa25d9ff4f12daea8cb79ee1b3ee81b9952. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #260, opened 2026-05-15
fix(nextjs): stop flagging response.headers and local Map/Set/Headers in GET handlers (#206)

### LINKED ISSUE #206: False positive: nextjs-no-side-effect-in-get-handler flags Response.headers.set() and similar Headers API calls

NisargIO:
## Summary

Rewrites `nextjs-no-side-effect-in-get-handler` precision so it no longer floods Next.js codebases with false positives from response-shaping calls, while still catching real CSRF-relevant server-state mutations.

Fixes #206. Supersedes #209, #211, #233, #238. The handler-resolution improvements from #251 (cron route skip + depth-bounded `const GET = withAuth(handler)` resolution) are also folded in; the broader rule expansion in #251 remains an independent PR.

### No longer flagged (false positives — were ~138 errors in one Next.js 14 codebase)

- `res.headers.set/append/delete(...)` and any chain ending in `.headers` (handles `NextResponse.json({...}).headers.set(...)`, `(await fetcher()).headers.append(...)`).
- Locally-constructed `new Map/Set/WeakMap/WeakSet/Headers/URLSearchParams/FormData/Response/NextResponse(...)` bindings and any mutation on those aliases.
- `new URL(...).searchParams.set(...)` and any `.searchParams.*()` chain.
- `headers()` / `(await headers())` from `next/headers` (returns `ReadonlyHeaders`; any mutation would throw at runtime) and any aliased `const h = headers(); h.get(...)`.
- Route handlers under `/cron/` or `/jobs/cron/` — Vercel Cron always invokes GET and is expected to do real work.

### Still flagged (real CSRF-relevant side effects)

- ORM mutations: drizzle `db.update(table).set({...}).where(...)`, `prisma.user.create(...)`, `db.insert(...)`, `repository.upsert(...)`.
- Module-level mutable state: `const cache = new Map()` declared outside the handler, then `cache.set(...)` inside.
- `fetch(url, { method: 'POST' | 'PUT' | 'DELETE' | 'PATCH' })`.
- `cookies().set/append/delete()` in **all** forms — direct, `(await cookies()).set(...)`, and (new) aliased `const cs = await cookies(); cs.set(...)` (the previously-missed aliased form).
- Mutating route segments (`/logout`, `/signout`, `/unsubscribe`, `/delete`, …).

## Implementation

`packages/oxlint-plugin-react-doctor/src/plugin/utils/find-side-effect.ts` now takes an options bag and short-circuits any mutation call whose receiver chain is structurally \"safe\" (a `.headers`/`.searchParams` access, a `new X()` of a safe constructor, a `Response`/`NextResponse` factory, or an Identifier in a per-handler safe-bindings set). The rule itself collects two binding sets before scanning:

- `collectLocallyScopedSafeBindings(handlerBody)` — `const X = new Map()`, `const X = new Headers()`, `const X = NextResponse.json(...)`, `const X = await headers()`, etc.
- `collectLocallyScopedCookieBindings(handlerBody)` — `const X = cookies()` and `const X = await cookies()`. Any `<alias>.set/append/delete()` is reported as `cookies().<method>()` so the diagnostic stays consistent.

The rule also gained depth-bounded handler-binding resolution (`GET_HANDLER_BINDING_RESOLUTION_DEPTH = 3`) so `export const GET = withAuth(handler)` and `export const GET = app.get('/x', handler)` chains get scanned correctly.

## Test plan

`packages/react-doctor/tests/regressions/nextjs-get-side-effects.test.ts` (new) builds 23 isolated synthetic projects covering:

- **13 false-negative cases** that must produce 0 hits, including the verbatim issue #206 repro and a 138x stress test that synthesises the original reported codebase.
- **9 true-positive cases** that must produce exactly 1 hit each, with the diagnostic naming the right operation (drizzle, prisma, module-level cache, mutating fetch, direct/awaited/aliased cookies, mutating route segment).

`packages/react-doctor/tests/run-oxlint/nextjs.test.ts` also gains three integration assertions on new fixtures under `tests/fixtures/nextjs-app/src/app/api/` to lock the behavior in the shared Next.js fixture project.

```bash
nr typecheck   # all packages pass
nr lint        # 0 warnings / 0 errors
nr test        # 725 / 725 pass (was 724 + 23 new = 725 wait let me recount — final: 78 files, 725 tests passing)
nr format      # no diff
```

## Suggested follow-ups

- Close #209, #211, #233, #238 as superseded.
- Land #251 separately once its rule-expansion surface is reviewed; the handler-resolution piece it shares with this PR is now on `main`.

## Eval results

Re-ran against the [larger sandbox corpus](https://github.com/millionco/react-doctor-evals) (500 repos, scanned via Vercel Sandbox in parallel) on `main` vs this branch:

| | `main` (8a3a1be) | this PR (71ecf0a) |
|---|---|---|
| Total diagnostics (all rules) | **66,276** | **66,276** |
| `nextjs-no-side-effect-in-get-handler` | 8 | 8 |
| Net diff in any rule | — | **0** |

**Zero functional changes across 500 repos.** The 51 line-level diffs the tool surfaces are 100% message-text encoding artifacts (same file/line/column/rule, just em-dash UTF-8 rendering); no diagnostic actually moves. All 8 `nextjs-no-side-effect-in-get-handler` hits are valid true positives in the corpus and remain flagged:

- `supabase/supabase:app/auth/confirm/route.ts:6` — `supabase.auth.verifyOtp(...)` in a GET handler (textbook CSRF risk on email confirm links)
- `unkeyed/unkey:app/auth/sso-callback/[[...sso-callback]]/route.ts:6`, `app/join/route.ts:9` — OAuth/invite acceptance with side effects
- `better-auth/better-auth:app/.well-known/oauth-*/route.ts` (4 hits) — discovery routes that mutate session/cache
- `umami-software/umami:src/app/api/websites/[websiteId]/event-data/route.ts:8` — event-data write inside GET

The targeted false positives (`response.headers.set`, local `new Map/Set/Headers`, `headers()` reads, aliased `cookies()`) and the new true positive (depth-bounded `export const GET = withAuth(handler)` resolution) seen on the earlier 14-repo set don't appear in this 500-repo slice — but the corpus-wide result is **0 regressions in any rule across 66,276 diagnostics**, plus the 23-case synthetic regression suite covering every pattern in the issue.

---

> [!NOTE]
> **Medium Risk**
> Medium risk: changes security-focused lint rules and their side-effect/auth-call detection logic, which could introduce false negatives/positives in real projects despite added regression coverage.
> 
> **Overview**
> **Improves security lint precision for server GET handlers and server actions.** `nextjs-no-side-effect-in-get-handler` is rewritten to avoid flagging response/header shaping and locally-created mutable objects (e.g. `response.headers.*`, `new Map/Set/Headers`, `URL.searchParams`) while still reporting real state changes (DB mutations, mutating `fetch` methods, cookie writes), and it now skips cron routes and resolves wrapped/aliased GET handlers via bounded binding resolution.
> 
> **Enhances shared side-effect detection and TanStack Start parity.** `findSideEffect` now accepts per-handler binding context to treat “safe” receiver chains as non-mutations and to correctly attribute aliased `cookies()` mutations; TanStack Start’s `tanstack-start-get-mutation` is updated to analyze the handler body with the same binding-aware logic.
> 
> **Extends `server-auth-actions` recognition and configurability.** The rule now detects auth checks in member calls and TS-wrapped/optional-chained callees, avoids counting nested-helper auth calls, supports additional server action export forms, and allows projects to provide `serverAuthFunctionNames`, which is plumbed from `react-doctor` config through core oxlint settings. Extensive new regression/integration tests and fixtures lock in these behaviors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit beadd1e92952fa35cb2a36a2d32a3129b80acfa7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.changeset/nextjs-get-side-effect-false-positives.md
```diff
@@ -0,0 +1,46 @@
...
+"react-doctor": patch
+"eslint-plugin-react-doctor": patch
+"oxlint-plugin-react-doctor": patch
+---
+
+Fix trust-breaking false positives in `nextjs-no-side-effect-in-get-handler`
+(issue #206). The rule used to flag any `<member>.<set|append|delete|create|
+insert|update|upsert|remove|destroy>()` call as a server-state mutation,
+which flooded one Next.js 14 codebase with 138 false errors — every single
+one a `response.headers.set(...)` response-shaping call.
```
aidenybai: no

aidenybai:
CI

NisargIO:
bugbot run

PR #261, opened 2026-05-15
fix(server-auth-actions): accept member-expression auth calls like `auth0.getSession()` (#239)

### LINKED ISSUE #239: False positives on auth gating server functions

NisargIO:
## Summary

Stacked on top of #260.

`server-auth-actions` only recognised auth calls when the callee was a bare `Identifier` — `auth()`, `getSession()`, `getUser()`, etc. Every member-expression flavour (`auth0.getSession()`, `supabase.auth.getSession()`, `clerkClient.getUser(userId)`, `request.auth()`) slipped through and the whole server action got flagged as missing an auth check. One reporter in #239 hit 139 false positives — essentially every server action in the repo.

Route every `CallExpression` through the existing [`getCalleeName`](packages/oxlint-plugin-react-doctor/src/plugin/utils/get-callee-name.ts) utility, which already resolves both `Identifier` and `MemberExpression` callees to a single name. `walkAst` already descends into `AwaitExpression` / `ChainExpression` children, so awaited (`await auth0.getSession()`) and optional-chained (`auth0?.getSession()`) variants fall out for free with no extra branching.

This is more elegant than [#240](https://github.com/millionco/react-doctor/pull/240) (which adds a second `else if` branch duplicating the identifier check) and reuses code we already have.

### Now accepted as an auth check

- `await auth0.getSession()` — the verbatim repro
- `auth0.getSession()` (non-awaited)
- `await supabase.auth.getSession()` (chained member expression)
- `await clerkClient.getUser(userId)`
- `await auth0?.getSession()` (optional chaining)
- `await auth()` / `await getSession()` (the original Identifier forms — still works)

### Still flagged

- Server actions with no auth-related call inside the first 10 statements.
- Server actions calling unrelated `.something()` member methods.
- Auth checks that happen after the lookahead window.

## Test plan

- [x] Added `packages/react-doctor/tests/regressions/server-auth-actions-member-call.test.ts` with 6 false-negative cases and 3 true-positive cases.
- [x] `nr typecheck`, `nr lint`, `nr test` (738/738), `nr format` all pass.
- [x] Existing `server-auth-actions` regressions in `packages/react-doctor/tests/run-oxlint/nextjs.test.ts` still pass.

Closes #239. Supersedes #240.

## Eval results

Re-ran against the [larger sandbox corpus](https://github.com/millionco/react-doctor-evals) (500 repos, scanned via Vercel Sandbox in parallel) on `main` vs parent #260 vs this branch:

| | `main` (8a3a1be) | parent #260 (71ecf0a) | this PR (3b9af6a) |
|---|---|---|---|
| Total diagnostics (all rules) | **66,276** | **66,276** | **66,276** |
| `server-auth-actions` | 54 | 54 | 54 |
| Net diff in any rule | — | **0** | **0** |

**Zero functional changes across 500 repos.** The 43 line-level diffs vs `main` are 100% message-text encoding artifacts (same file/line/column/rule, em-dash rendering); no diagnostic actually moves. All 54 `server-auth-actions` hits stay flagged — these are real true positives in `supabase/supabase`, `payloadcms/payload`, `unkeyed/unkey`, and others: server actions performing `prisma.*` / form-submit / database mutations with no auth call in the lookahead window.

The patterns this PR unblocks (`auth0.getSession()`, `supabase.auth.getSession()`, `clerkClient.getUser()`, optional-chained and awaited variants) don't appear in the 500-repo sample as actual *missed* false positives, so the fix is exercised by the new `server-auth-actions-member-call.test.ts` regression suite (6 false-negative + 3 true-positive cases) and the verbatim repro from #239. Corpus-wide result: **0 regressions in any rule across 66,276 diagnostics**.

---

> [!NOTE]
> **Medium Risk**
> Updates the `server-auth-actions` security rule’s AST matching and adds config-driven behavior, which can change which server actions are flagged (false positives/negatives) but does not affect runtime application code.
> 
> **Overview**
> Fixes `server-auth-actions` so top-of-action auth checks are recognized when invoked via member/chain/optional-chain or TS-wrapped callees (e.g. `auth0.getSession()`, `supabase.auth.getSession()`, `auth!()`, `(auth as T)()`), and expands what export shapes are inspected (named exports, `export const ... = async ...`, and `export default async function`). It also tightens detection by **not** counting auth calls inside nested/unused helpers and by disambiguating overly-generic method names like `.getUser()` using an auth-ish receiver heuristic.
> 
> Adds a new user config key `serverAuthFunctionNames` that is forwarded from `react-doctor` config into oxlint `settings` and treated as an allowlist of custom auth guards. Introduces shared `getReactDoctor*Setting` helpers for safe typed reads from `context.settings`, updates `no-secrets-in-client-code` to use them, documents the new config in the README/types, and adds regression tests covering the new `server-auth-actions` behaviors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fcd7f8b3e04d1fa2145cd0370aeaf8312ce95ce5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.changeset/server-auth-actions-member-call.md
```diff
@@ -0,0 +1,27 @@
+---
```
aidenybai: remove

PR #264, opened 2026-05-16
fix(server-auth-actions): accept member-expression auth calls like `auth0.getSession()` (#239)

### LINKED ISSUE #239: False positives on auth gating server functions

NisargIO:
## Summary

Re-opens [#261](https://github.com/millionco/react-doctor/pull/261), which GitHub auto-closed after a force-push during rebase. Stacked on top of #263.

`server-auth-actions` only recognised auth calls when the callee was a bare `Identifier` — `auth()`, `getSession()`, `getUser()`, etc. Every member-expression flavour (`auth0.getSession()`, `supabase.auth.getSession()`, `clerkClient.getUser(userId)`, `request.auth()`) slipped through and the whole server action got flagged as missing an auth check. One reporter in #239 hit 139 false positives — essentially every server action in the repo.

Route every `CallExpression` through the existing [`getCalleeName`](packages/oxlint-plugin-react-doctor/src/plugin/utils/get-callee-name.ts) utility, which already resolves both `Identifier` and `MemberExpression` callees to a single name. `walkAst` already descends into `AwaitExpression` / `ChainExpression` children, so awaited (`await auth0.getSession()`) and optional-chained (`auth0?.getSession()`) variants fall out for free with no extra branching.

This is more elegant than [#240](https://github.com/millionco/react-doctor/pull/240) (which adds a second `else if` branch duplicating the identifier check) and reuses code we already have.

### Now accepted as an auth check

- `await auth0.getSession()` — the verbatim repro
- `auth0.getSession()` (non-awaited)
- `await supabase.auth.getSession()` (chained member expression)
- `await clerkClient.getUser(userId)`
- `await auth0?.getSession()` (optional chaining)
- `await auth()` / `await getSession()` (the original Identifier forms — still works)

### Still flagged

- Server actions with no auth-related call inside the first 10 statements.
- Server actions calling unrelated `.something()` member methods.
- Auth checks that happen after the lookahead window.

## Test plan

- [x] Added `packages/react-doctor/tests/regressions/server-auth-actions-member-call.test.ts` with 6 false-negative cases and 3 true-positive cases.
- [x] `nr typecheck`, `nr lint`, `nr test` (918/918), `nr format:check` all pass.
- [x] Existing `server-auth-actions` regressions in `packages/react-doctor/tests/run-oxlint/nextjs.test.ts` still pass.

Closes #239. Supersedes #240.


NisargIO:
Closing in favor of reopened #261 (same diff, same branch).

PR #471, opened 2026-05-25
[MIL-1547] Recognize `subscription.remove()` as effect cleanup

rayhanadev:
## Why

Fixes false positives where `effect-needs-cleanup` reported effects that already return a valid cleanup, such as React Native `subscription.remove()` or a returned `stopInterval` helper.

This is not just adding `.remove()` to a global allowlist. A global `.remove()` rule would hide real leaks because many unrelated objects have a `remove()` method:

```tsx
useEffect(() => {
  window.addEventListener("resize", onResize);
  return () => {
    node.remove();
  };
}, []);
```

That cleanup removes `node`, not the resize listener, so the rule should still report it.

The safe version needs a small amount of binding analysis: accept `.remove()` only when the receiver is the value returned by the subscription call in the same effect.

```tsx
useEffect(() => {
  const subscription = AppState.addEventListener("change", onChange);
  return () => {
    subscription.remove();
  };
}, []);
```

React reruns effect cleanup before the next effect run and on unmount. A listener, subscription, interval, or timeout that is not cleaned up can leave stale callbacks behind. This PR keeps reporting those leaks while accepting library-supported cleanup shapes that can be found from the effect body and returned cleanup body.

## What changed

- Splits cleanup methods into global cleanup calls, cleanup-returning subscribe calls, and methods that are valid only on subscription-bound receivers.
- Accepts subscription-bound cleanup methods such as `subscription.remove()` and `simulation.stop()` only when the receiver comes from a subscribe-like call in the same effect.
- Accepts common indirect cleanup forms found in the eval: returned local cleanup functions, optional cleanup calls like `remove?.()`, iterator callback cleanup, and d3-style `.on(name, null)` listener removal.
- Keeps generic cleanup verbs narrow so unrelated calls like `node.remove()` do not hide real missing cleanup.
- Adds adversarial regression coverage for valid cleanup shapes and for false-negative traps where cleanup is unrelated or scoped inside an inner callback.

## Test plan

```sh
nr build # in packages/oxlint-plugin-react-doctor
nr typecheck # in packages/oxlint-plugin-react-doctor
nr test tests/regressions/state-rules/effect-needs-cleanup.test.ts tests/regressions/state-rules/prefer-use-sync-external-store.test.ts # in packages/react-doctor
nr format:check
git diff --check
```

Focused eval:

- Rule: `react-doctor/effect-needs-cleanup`
- Dataset: first 100 entries from `react-doctor-evals` `repos.json`
- Result: 26 diagnostics across 9 roots -> 20 diagnostics across 7 roots
- Scan failures: 0
- Remaining non-timer diagnostic was manually checked and appears to be a real missing `textEditor.off(...)` cleanup in tldraw.

---

> [!NOTE]
> **Medium Risk**
> Changes static analysis heuristics for two user-facing rules; incorrect binding logic could hide real leaks or reintroduce noise, though coverage is heavily regression-tested.
> 
> **Overview**
> Tightens **effect cleanup** detection for `effect-needs-cleanup` and aligned `prefer-use-sync-external-store` logic so fewer real cleanups are reported as missing.
> 
> Release method constants are split into **global** teardown (`removeEventListener`, `abort`, …), **cleanup-returning** subscribe calls (`subscribe` / `sub` only), and **receiver-bound** verbs (`remove`, `stop`, `dispose`, …) that count only when the object is bound from a subscribe-like call in the same effect—so patterns like React Native `sub.remove()` pass while unrelated `node.remove()` after `addEventListener` still fails.
> 
> `effect-needs-cleanup` now **collects cleanup bindings** (subscriptions, inferred cleanup functions from declarations/assignments) and feeds them into shared **`isCleanupReturn` / `isCleanupFunctionLike`** checks, including iterator-callback cleanup, optional `remove?.()`, d3-style `.on(name, null)`, and stricter rules for expression-body arrows (`addEventListener` no longer treated like `subscribe`). Large regression tests cover valid and adversarial shapes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d9514445e4f6b28ebb321837b39da906652af503. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/oxlint-plugin-react-doctor/src/plugin/constants/react.ts
+++ b/packages/oxlint-plugin-react-doctor/src/plugin/constants/react.ts
@@ -91,12 +91,8 @@ export const SUBSCRIPTION_METHOD_NAMES = new Set([
 // via `{ signal }`: a single `controller.abort()` removes every
 // listener bound to that signal in one shot, so it IS the matching
 // "remove" call even when no literal `removeEventListener(...)` is present.
-// `remove` covers React Native's EventSubscription objects returned by
-// APIs like AppState.addEventListener, Keyboard.addListener, Appearance,
-// and Dimensions.
 export const UNSUBSCRIPTION_METHOD_NAMES = new Set([
   "unsubscribe",
-  "remove",
   "removeEventListener",
   "removeListener",
   "off",
--- a/packages/oxlint-plugin-react-doctor/src/plugin/constants/react.ts
+++ b/packages/oxlint-plugin-react-doctor/src/plugin/constants/react.ts
@@ -81,17 +81,16 @@ export const SUBSCRIPTION_METHOD_NAMES = new Set([
   "sub",
 ]);
 
-// Methods that pair with the subscription methods above as their cleanup
-// counterparts. Used to recognize a valid `return () => removeEventListener(...)`
-// cleanup form even when the subscribe call is `addEventListener` rather
-// than a `subscribe()` whose return value gets re-bound.
+// Cleanup method names that are meaningful without knowing where the receiver
+// came from. A call like `target.removeEventListener(...)`, `emitter.off(...)`,
+// or `controller.abort()` describes the teardown in the method name itself, so
+// it is safe to accept on any receiver inside an effect cleanup.
 //
-// `abort` covers the AbortController pattern, which is the recommended
-// modern shape for tearing down `addEventListener` registrations bound
-// via `{ signal }`: a single `controller.abort()` removes every
-// listener bound to that signal in one shot, so it IS the matching
-// "remove" call even when no literal `removeEventListener(...)` is present.
-export const UNSUBSCRIPTION_METHOD_NAMES = new Set([
+// Keep generic resource verbs such as `remove`, `dispose`, `destroy`, and
+// `teardown` OUT of this set. On an arbitrary receiver, `node.remove()` or
+// `cache.dispose()` does not prove it releases the listener/subscription that
+// was created earlier in the effect.
+export const GLOBAL_RELEASE_METHOD_NAMES = new Set([
   "unsubscribe",
   "removeEventListener",
   "removeListener",
@@ -102,11 +101,17 @@ export const UNSUBSCRIPTION_METHOD_NAMES = new Set([
   "abort",
 ]);
 
-// Methods accepted only when called on a variable bound to a subscribe-like
-// return value (`const sub = source.addListener(...); return () => sub.remove()`).
-// These names are too generic to trust on arbitrary receivers, but they are
-// common release verbs for subscription/resource objects.
-export const BOUND_SUBSCRIPTION_RELEASE_METHOD_NAMES = new Set([
+// Cleanup method names that are only meaningful when the receiver is the value
+// returned by a subscribe-like call in the same effect. For example:
+//
+//   const sub = AppState.addEventListener(...);
+//   return () => sub.remove();
+//
+// In that shape, the receiver (`sub`) ties the generic verb (`remove`) back to
+// the resource created by the effect. Without that binding relationship, these
+// verbs are too broad and would create false negatives for unrelated cleanup
+// calls.
+export const BOUND_RESOURCE_RELEASE_METHOD_NAMES = new Set([
   "remove",
   "unsubscribe",
   "unsub",
@@ -119,11 +124,11 @@ export const BOUND_SUBSCRIPTION_RELEASE_METHOD_NAMES = new Set([
 // Identifier names recognized as "this is a release/teardown call"
 // when they appear as a direct call inside an effect's cleanup
 // return — covers both library unsubscribe shorthands
-// (UNSUBSCRIPTION_METHOD_NAMES) and the generic teardown vocabulary
+// (GLOBAL_RELEASE_METHOD_NAMES) and the generic teardown vocabulary
 // (`cleanup`, `dispose`, `destroy`, `teardown`). Matched
 // case-insensitively at the call site.
 export const CLEANUP_LIKE_RELEASE_CALLEE_NAMES = new Set([
-  ...UNSUBSCRIPTION_METHOD_NAMES,
+  ...GLOBAL_RELEASE_METHOD_NAMES,
   "cleanup",
   "dispose",
   "destroy",
--- a/packages/oxlint-plugin-react-doctor/src/plugin/constants/react.ts
+++ b/packages/oxlint-plugin-react-doctor/src/plugin/constants/react.ts
@@ -102,6 +102,20 @@ export const UNSUBSCRIPTION_METHOD_NAMES = new Set([
   "abort",
 ]);
 
+// Methods accepted only when called on a variable bound to a subscribe-like
+// return value (`const sub = source.addListener(...); return () => sub.remove()`).
+// These names are too generic to trust on arbitrary receivers, but they are
+// common release verbs for subscription/resource objects.
+export const BOUND_SUBSCRIPTION_RELEASE_METHOD_NAMES = new Set([
+  "remove",
+  "unsubscribe",
+  "unsub",
+  "cleanup",
+  "dispose",
+  "destroy",
+  "teardown",
+]);
+
 // Identifier names recognized as "this is a release/teardown call"
 // when they appear as a direct call inside an effect's cleanup
 // return — covers both library unsubscribe shorthands
--- a/packages/oxlint-plugin-react-doctor/src/plugin/constants/react.ts
+++ b/packages/oxlint-plugin-react-doctor/src/plugin/constants/react.ts
@@ -81,22 +81,12 @@ export const SUBSCRIPTION_METHOD_NAMES = new Set([
   "sub",
 ]);
 
-// Subscribe-like methods whose return value is itself the cleanup function.
-// Do not include EventTarget/EventEmitter-style registrations here:
-// `addEventListener`, `addListener`, and `on` often return void or a
-// subscription object, so accepting their return value as cleanup would hide
-// missing teardown.
+// Subscribe-like methods that return the cleanup function directly.
+// EventTarget/EventEmitter APIs often return void or subscription objects.
 export const CLEANUP_RETURNING_SUBSCRIPTION_METHOD_NAMES = new Set(["subscribe", "sub"]);
 
-// Cleanup method names that are meaningful without knowing where the receiver
-// came from. A call like `target.removeEventListener(...)`, `emitter.off(...)`,
-// or `controller.abort()` describes the teardown in the method name itself, so
-// it is safe to accept on any receiver inside an effect cleanup.
-//
-// Keep generic resource verbs such as `remove`, `dispose`, `destroy`, and
-// `teardown` OUT of this set. On an arbitrary receiver, `node.remove()` or
-// `cache.dispose()` does not prove it releases the listener/subscription that
-// was created earlier in the effect.
+// Cleanup methods that are specific enough to trust on any receiver.
+// Generic verbs like `remove` need a known subscription receiver.
 export const GLOBAL_RELEASE_METHOD_NAMES = new Set([
   "unsubscribe",
   "removeEventListener",
@@ -108,16 +98,7 @@ export const GLOBAL_RELEASE_METHOD_NAMES = new Set([
   "abort",
 ]);
 
-// Cleanup method names that are only meaningful when the receiver is the value
-// returned by a subscribe-like call in the same effect. For example:
-//
-//   const sub = AppState.addEventListener(...);
-//   return () => sub.remove();
-//
-// In that shape, the receiver (`sub`) ties the generic verb (`remove`) back to
-// the resource created by the effect. Without that binding relationship, these
-// verbs are too broad and would create false negatives for unrelated cleanup
-// calls.
+// Generic cleanup methods accepted only on values returned by subscribe-like calls.
 export const BOUND_RESOURCE_RELEASE_METHOD_NAMES = new Set([
   "remove",
   "cleanup",
@@ -127,12 +108,7 @@ export const BOUND_RESOURCE_RELEASE_METHOD_NAMES = new Set([
   "teardown",
 ]);
 
-// Identifier names recognized as "this is a release/teardown call"
-// when they appear as a direct call inside an effect's cleanup
-// return — covers both library unsubscribe shorthands
-// (GLOBAL_RELEASE_METHOD_NAMES) and the generic teardown vocabulary
-// (`cleanup`, `dispose`, `destroy`, `teardown`). Matched
-// case-insensitively at the call site.
+// Direct cleanup function names accepted inside returned cleanup functions.
 export const CLEANUP_LIKE_RELEASE_CALLEE_NAMES = new Set([
   ...GLOBAL_RELEASE_METHOD_NAMES,
   "cleanup",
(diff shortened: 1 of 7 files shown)


aidenybai (commented):
ran the suite locally (effect-needs-cleanup 20/20, prefer-use-sync-external-store 8/8, tsc clean) and probed a few edge cases. direction is right. a handful of inline notes below, mostly small.

packages/oxlint-plugin-react-doctor/src/plugin/constants/react.ts
```diff
@@ -92,14 +91,34 @@ export const UNSUBSCRIPTION_METHOD_NAMES = new Set([
...
+// calls.
+export const BOUND_RESOURCE_RELEASE_METHOD_NAMES = new Set([
+  "remove",
+  "unsubscribe",
+  "unsub",
+  "cleanup",
+  "dispose",
+  "destroy",
+  "teardown",
+]);
```
aidenybai, line 104: `unsubscribe` and `unsub` are already in `GLOBAL_RELEASE_METHOD_NAMES`, and `isReleaseLikeCall` falls through to the global check whenever the bound check fails. so the bound entries are unreachable. drop them here or add a note justifying the duplication.

packages/oxlint-plugin-react-doctor/src/plugin/rules/state-and-effects/effect-needs-cleanup.ts
```diff
@@ -146,7 +155,7 @@ const effectHasCleanupRelease = (callback: EsTreeNode): boolean => {
   if (!isNodeOfType(callback.body, "BlockStatement")) {
     return isSubscribeLikeCallExpression(callback.body);
   }
-  const knownBoundReleaseNames = collectReleasableBindingNames(callback);
+  const knownBoundReleaseNames = collectReleasableBindings(callback);
```
aidenybai: name now misrepresents the value. after the refactor this holds the whole `ReleasableBindings` struct (release names + subscription names), not just release names. rename to `releasableBindings` to match what it returns.

packages/oxlint-plugin-react-doctor/src/plugin/rules/state-and-effects/effect-needs-cleanup.ts
```diff
@@ -90,40 +90,49 @@ const findSubscribeLikeUsages = (callback: EsTreeNode): SubscribeLikeUsage[] =>
...
       }
       if (
         isNodeOfType(init.callee, "Identifier") &&
         TIMER_CALLEE_NAMES_REQUIRING_CLEANUP.has(init.callee.name)
       ) {
-        releasableNames.add(declarator.id.name);
+        bindings.releaseNames.add(declarator.id.name);
       }
     }
   }
```
aidenybai: this only walks `effectCallback.body.body`, so any `const sub = source.addListener(...)` declared inside an `if` block is invisible. the cleanup-return scan already walks nested return statements (see the existing `cleanup nested inside an if block` regression), so the bound-resource path is asymmetric and false-positives the gated-subscription shape:

```tsx
useEffect(() => {
  if (isOn) {
    const sub = source.addListener(handler);
    return () => sub.remove();
  }
}, [isOn]);
```

not a regression (pre-pr this case also failed) but worth a follow-up so `removeEventListener` and `sub.remove()` shapes behave the same in conditional cleanup.

packages/oxlint-plugin-react-doctor/src/plugin/rules/state-and-effects/utils/is-cleanup-return.ts
```diff
@@ -21,19 +23,27 @@ const isReleaseLikeCall = (
     return false;
   }
   if (isNodeOfType(callee, "MemberExpression") && isNodeOfType(callee.property, "Identifier")) {
-    return UNSUBSCRIPTION_METHOD_NAMES.has(callee.property.name);
+    if (
+      BOUND_RESOURCE_RELEASE_METHOD_NAMES.has(callee.property.name) &&
+      isNodeOfType(callee.object, "Identifier") &&
+      knownBoundSubscriptionNames.has(callee.object.name)
```
aidenybai, line 72: intentionally narrow but worth a `// TODO:` so the next person knows the bare-identifier restriction is deliberate. probes confirm `subRef.current.remove()`, `this.sub.remove()`, and `sub?.remove()` all false-positive today.

packages/oxlint-plugin-react-doctor/src/plugin/rules/state-and-effects/utils/is-cleanup-return.ts
```diff
@@ -44,6 +54,7 @@ const containsReleaseLikeCall = (
 export const isCleanupReturn = (
   returnedValue: EsTreeNode | null | undefined,
   knownBoundReleaseNames: ReadonlySet<string>,
+  knownBoundSubscriptionNames: ReadonlySet<string> = new Set(),
```
aidenybai: no caller relies on this default. drop it so the api is required-explicit. otherwise a future call site can silently fall back to an empty set and lose bound-resource detection without anyone noticing.

packages/oxlint-plugin-react-doctor/src/plugin/rules/state-and-effects/prefer-use-sync-external-store.ts
```diff
@@ -143,7 +143,7 @@ const cleanupReleasesSubscription = (
   if (!isNodeOfType(lastStatement, "ReturnStatement")) return false;
   const knownBoundReleaseNames = new Set<string>();
   if (boundUnsubscribeName) knownBoundReleaseNames.add(boundUnsubscribeName);
-  return isCleanupReturn(lastStatement.argument, knownBoundReleaseNames);
+  return isCleanupReturn(lastStatement.argument, knownBoundReleaseNames, knownBoundReleaseNames);
```
aidenybai: this change silently broadens what counts as cleanup for this rule too. a useState/useEffect reimplementation whose cleanup is `() => sub.remove()` will now also trip the diagnostic. add a positive regression so the cross-rule consequence is locked in (the test file only got a comment update).

aidenybai:
bugbot run

packages/oxlint-plugin-react-doctor/src/plugin/rules/state-and-effects/utils/is-cleanup-return.ts
```diff
@@ -1,60 +1,122 @@
...
 
+const isSynchronousIterationCallback = (node: EsTreeNode): boolean => {
+  const parentNode = node.parent;
+  if (!isNodeOfType(parentNode, "CallExpression")) return false;
+  if (!parentNode.arguments?.some((argument) => argument === node)) return false;
+  const callee = unwrapChainExpression(parentNode.callee);
+  return (
+    isNodeOfType(callee, "MemberExpression") &&
+    isNodeOfType(callee.property, "Identifier") &&
+    SYNC_ITERATION_METHOD_NAMES.has(callee.property.name)
```
aidenybai: SYNC_ITERATION could technically include all of the Array methods right? .e.g .map technically iterates

packages/react-doctor/tests/regressions/state-rules/effect-needs-cleanup.test.ts
```diff
@@ -135,13 +255,231 @@ export const Clock = () => {
...
+  });
+
   it("does NOT flag expression-body arrow whose subscribe return is the implicit cleanup (Bugbot #157)", async () => {
-    // Regression: \`useEffect(() => store.subscribe(handler), [])\` is a
-    // common compact form — the arrow's expression body IS the body,
-    // and the subscribe call's return value (the unsubscribe fn) is
-    // implicitly returned as the effect's cleanup. The earlier
-    // detector rejected non-BlockStatement bodies outright and
-    // false-positived this shape.
+    // Expression-body effects return the subscribe cleanup directly.
```
aidenybai: delete

packages/react-doctor/tests/regressions/state-rules/effect-needs-cleanup.test.ts
```diff
@@ -161,11 +499,27 @@ export const Subscribe = () => {
...
+    const hits = await collectRuleHits(projectDir, "effect-needs-cleanup");
+    expect(hits).toHaveLength(1);
+  });
+
   it("does NOT flag a `setTimeout` that lives inside the cleanup return (Bugbot #157 round 3)", async () => {
-    // Regression: the subscribe/timer scanner walked the entire
-    // callback including the cleanup return body. A \`setTimeout\` in
-    // the cleanup is a disposal step, not a new registration; it
-    // should not produce a 'missing cleanup' diagnostic.
+    // Timers inside returned cleanup functions are disposal work.
```
aidenybai: delete

packages/react-doctor/tests/regressions/state-rules/effect-needs-cleanup.test.ts
```diff
@@ -191,11 +545,7 @@ export const Beacon = () => {
   });
 
   it("does NOT flag `return () => unsub()` after `const unsub = subscribe(...)` (Bugbot #157 round 3)", async () => {
-    // Regression: the Identifier-callee cleanup regex only matched
-    // long-form names (unsubscribe / cleanup / dispose / destroy /
-    // teardown). \`unsub\` (and other short forms) were missing,
-    // producing a false positive on the canonical bind-the-result-
-    // and-call-it shape.
+    // The cleanup name comes from the subscribe result, not a naming heuristic.
```
aidenybai: delete

packages/react-doctor/tests/regressions/state-rules/effect-needs-cleanup.test.ts
```diff
@@ -219,10 +569,7 @@ export const Subscribe = () => {
   });
 
   it("recognizes the generic teardown vocabulary (`cleanup`, `dispose`, `destroy`, `teardown`) as a release call", async () => {
-    // The release-callee allowlist now lives in `constants.ts` as
-    // `CLEANUP_LIKE_RELEASE_CALLEE_NAMES`. Each of the generic
-    // teardown verbs satisfies the cleanup check on its own — no
-    // false positive on this shape.
+    // Generic teardown names are valid as direct cleanup calls.
```
aidenybai: delete

packages/react-doctor/tests/regressions/state-rules/effect-needs-cleanup.test.ts
```diff
@@ -274,10 +621,55 @@ export const Subscribe = () => {
...
+`,
+        },
+      },
+    );
+
+    const hits = await collectRuleHits(projectDir, "effect-needs-cleanup");
+    expect(hits).toHaveLength(1);
+  });
+
+  // Guarded effects often return cleanup from inside a branch.
```
aidenybai: delete

packages/react-doctor/tests/regressions/state-rules/effect-needs-cleanup.test.ts
```diff
@@ -353,11 +745,7 @@ export const Subscribe = () => {
     expect(hits).toHaveLength(0);
   });
 
-  // Regression for #310: AbortController is the modern, idiomatic way to
-  // tear down many `addEventListener` registrations in one call. A single
-  // `controller.abort()` removes every listener that was bound via
-  // `{ signal: controller.signal }`, so the cleanup return IS valid even
-  // though no literal `removeEventListener(...)` appears.
+  // AbortController can clean up listeners registered with its signal.
```
aidenybai: delete

packages/oxlint-plugin-react-doctor/src/plugin/rules/state-and-effects/effect-needs-cleanup.ts
```diff
@@ -90,40 +71,93 @@ const findSubscribeLikeUsages = (callback: EsTreeNode): SubscribeLikeUsage[] =>
...
-        releasableNames.add(declarator.id.name);
+        bindings.cleanupFunctionNames.add(declarator.id.name);
       }
     }
-  }
-  return releasableNames;
+  });
+
+  walkAst(effectCallback.body, (child: EsTreeNode) => {
+    // Local helpers may assign cleanup into an effect-scope variable.
```
aidenybai: del

packages/oxlint-plugin-react-doctor/src/plugin/rules/state-and-effects/effect-needs-cleanup.ts
```diff
@@ -90,40 +71,93 @@ const findSubscribeLikeUsages = (callback: EsTreeNode): SubscribeLikeUsage[] =>
...
-// previous "any Identifier is fine" behavior.
-const collectReleasableBindingNames = (effectCallback: EsTreeNode): Set<string> => {
-  const releasableNames = new Set<string>();
+interface CleanupBindings {
+  cleanupFunctionNames: Set<string>;
+  subscriptionNames: Set<string>;
+  effectScopeVariableNames: Set<string>;
+}
+
+// Trust cleanup names from the effect scope. Also allow local helpers to assign
```
aidenybai: delete

packages/oxlint-plugin-react-doctor/src/plugin/constants/react.ts
```diff
@@ -102,14 +98,19 @@ export const UNSUBSCRIPTION_METHOD_NAMES = new Set([
...
+export const BOUND_RESOURCE_RELEASE_METHOD_NAMES = new Set([
+  "remove",
+  "cleanup",
+  "dispose",
+  "destroy",
+  "stop",
+  "teardown",
+]);
+
+// Direct cleanup function names accepted inside returned cleanup functions.
```
aidenybai: delete

packages/oxlint-plugin-react-doctor/src/plugin/rules/state-and-effects/utils/is-cleanup-return.ts
```diff
@@ -1,60 +1,148 @@
...
+const unwrapChainExpression = (node: EsTreeNode): EsTreeNode =>
+  isNodeOfType(node, "ChainExpression") ? node.expression : node;
+
+const isNullLiteral = (node: EsTreeNode | null | undefined): boolean =>
+  isNodeOfType(node, "Literal") && node.value === null;
+
+const isListenerRemovalViaNullHandler = (callNode: EsTreeNode): boolean => {
+  if (!isNodeOfType(callNode, "CallExpression")) return false;
+  const callee = unwrapChainExpression(callNode.callee);
+  // d3-style `.on(name, null)` removes a listener.
```
aidenybai, line 42: good

packages/oxlint-plugin-react-doctor/src/plugin/rules/state-and-effects/effect-needs-cleanup.ts
```diff
@@ -90,40 +71,93 @@ const findSubscribeLikeUsages = (callback: EsTreeNode): SubscribeLikeUsage[] =>
...
+        bindings.subscriptionNames.add(bindingName);
+        if (isCleanupReturningSubscribeLikeCallExpression(init)) {
+          bindings.cleanupFunctionNames.add(bindingName);
+        }
       }
+    }
+  });
+
+  walkAst(effectCallback.body, (child: EsTreeNode) => {
+    // Function expressions create inner scopes; their cleanup is not returned
```
aidenybai: no

PR #472, opened 2026-05-25
Fix Expo config plugin reachability

### LINKED ISSUE #468: False positive unused-file for Expo local config plugins referenced by string path
### Summary

React Review reports local Expo config plugins as `unused-file` when they are referenced from `app.config.ts` via Expo's documented string-path plugin syntax.

### What happened

In an Expo SDK 56 app, `app.config.ts` uses local config plugins like this:

```ts
export default (): ExpoConfig => ({
  plugins: [
    "./plugins/android-secure-flag.plugin.ts",
    "./plugins/android-faster-startup-no-bundle.ts",
    "./plugins/android-day-night-theme.ts",
    "./plugins/android-res-config.ts",
  ],
})
```

React Review flagged each target plugin file:

```text
plugins/android-day-night-theme.ts:0
Unused file — not reachable from any entry point
Rule: unused-file

plugins/android-faster-startup-no-bundle.ts:0
Unused file — not reachable from any entry point
Rule: unused-file

plugins/android-res-config.ts:0
Unused file — not reachable from any entry point
Rule: unused-file

plugins/android-secure-flag.plugin.ts:0
Unused file — not reachable from any entry point
Rule: unused-file
```

These files are not unused from Expo's perspective: Expo resolves and runs local config plugins listed by string path in the `plugins` array.

### Why this seems like a false positive

Expo docs show local config plugins being added by path in app config:

```ts
plugins: [
  ["./plugins/withPlugin.ts"],
]
```

The app validates successfully:

```sh
npx expo config --json
npx expo config --type introspect --json
```

Native builds also pass:

```sh
cd android && ./gradlew :app:assembleDebug
bun ios -- --device generic --no-bundler --no-install
```

### Expected behavior

`unused-file` should treat local Expo config plugin string paths in `app.config.ts` / `app.config.js` / `app.json` as reachability edges, so the plugin implementation files are not reported as unused.

### Notes

I searched existing issues for combinations of `unused-file`, `Expo`, `config plugin`, and `app.config.ts` and did not find an existing report.

Interestingly, running the local CLI did not reproduce the exact `unused-file` warnings for me:

```sh
npx react-doctor@latest --verbose --diff --offline --fail-on none
```

This returned only unrelated warnings in changed React UI files, while React Review in the PR reported the four `unused-file` warnings above. So this may be specific to the React Review environment or to the dead-code/reachability pass used there.

rayhanadev:
## Summary
- Add a `reachability` entry resolver registry for framework-specific roots that static imports cannot reveal.
- Add an Expo config plugin resolver that treats documented local plugin string paths in `app.config.*` and `app.json` as reachable files.
- Wire those extra entry patterns into the existing `deslop-js` dead-code scan without changing diagnostic mapping or ignore behavior.

## Before / After
Before this change, Expo apps could reference local config plugins with Expo's documented string-path syntax, but React Doctor's `unused-file` pass did not treat those strings as graph edges:

```ts
export default () => ({
  plugins: [
    "./plugins/android-secure-flag.plugin.ts",
    ["./plugins/android-day-night-theme", { enabled: true }],
  ],
});
```

Those plugin files could be reported as:

```text
plugins/android-secure-flag.plugin.ts:0
Unused file — not reachable from any entry point
Rule: unused-file
```

After this change, the reachability registry discovers those local Expo plugin paths before `deslop-js` runs, adds them as entry roots, and the same plugin files are no longer reported as unused. Package plugin names like `expo-camera` are still ignored as file paths.

## How It Works
React Doctor still delegates whole-project dead-code analysis to `deslop-js`. The new `packages/core/src/reachability/` layer runs just before `analyze()` and contributes additional `entryPatterns` for files that are reachable by framework convention rather than JavaScript imports.

`checkDeadCode()` now builds the normal deslop config, calls `collectReachabilityEntryPatterns({ rootDirectory, project })`, and merges any returned paths into `config.entryPatterns`. Everything after that stays inside the existing deslop reachability graph and diagnostic conversion path.

The registry shape is intentionally small:
- `ReachabilityEntryResolver` defines `id`, `isEnabled()`, and `collectEntryPatterns()`.
- `reachabilityEntryResolvers` is the list of framework-specific resolvers.
- `collectReachabilityEntryPatterns()` runs enabled resolvers and dedupes the returned entry paths.

## Expo Resolver Details
The Expo resolver is enabled for Expo, React Native, or RN-containing workspaces using the existing `ProjectInfo` signals.

It scans for Expo app config files:
- `app.config.ts`, `.mts`, `.cts`, `.js`, `.mjs`, `.cjs`
- `app.json`

It recognizes Expo's documented plugin forms:
- `plugins: ["./plugins/withPlugin.ts"]`
- `plugins: [["./plugins/withPlugin", { ...options }]]`
- `app.json` / `expo.plugins` with the same string or tuple forms

Only local relative paths (`./` or `../`) are treated as files. Package plugin names such as `expo-camera` and `expo-router` are intentionally ignored, so package-name lookalikes do not become false entry roots. Resolved local paths are returned relative to the project root, with extension and `index.*` resolution for common JS/TS extensions.

The resolver uses Expo's documented behavior as prior art: https://docs.expo.dev/config-plugins/plugins/

## Why This Shape
This avoids embedding Expo-specific parsing directly in `checkDeadCode()`. The dead-code adapter remains responsible for config/ignore forwarding and diagnostic mapping, while convention-based graph roots live behind a named reachability registry. Future framework conventions can add resolvers without widening `checkDeadCode()`.

Longer term, this kind of framework-specific entry extraction could move upstream into `deslop-js`, which already has similar config-string and framework entry discovery patterns.

## Test Plan
- `pnpm --filter @react-doctor/core typecheck`
- `pnpm --dir packages/core exec vp test run tests/reachability/collect-entry-patterns.test.ts`
- `pnpm --filter @react-doctor/core build`
- `pnpm --dir packages/react-doctor exec vp test run tests/check-dead-code.test.ts`
- `pnpm format:check`


rayhanadev:
Closes #468

aidenybai:
i think this should be in deslop-js @rayhanadev

aidenybai (changes requested):
check if this should be in deslop-js

rayhanadev:
yeah deslop probably makes more sense, will pr over there instead :+1:

PR #480, opened 2026-05-25
[MIL-1548] Fix bottom sheet React Native false positive

rayhanadev:
## Why

Catches legacy JS-thread bottom sheet packages without flagging `@gorhom/bottom-sheet`.

The old rule reported Gorhom imports as JS-implemented, but modern `@gorhom/bottom-sheet` is built on `react-native-reanimated` and `react-native-gesture-handler`, so that warning was a false positive. The old copy also overstated `<Modal presentationStyle="formSheet">`: it can fit narrow iOS form-sheet flows, but it is not a general replacement for bottom sheets with snap points, scrollables, backdrops, or programmatic snap control.

Before:

```ts
import BottomSheet from "@gorhom/bottom-sheet";
```

React Doctor reported `rn-bottom-sheet-prefer-native` and recommended `formSheet` as if it were a full bottom-sheet replacement.

After:

```ts
import BottomSheet from "@gorhom/bottom-sheet";
```

React Doctor does not report `rn-bottom-sheet-prefer-native` for Gorhom imports. Legacy packages such as `react-native-raw-bottom-sheet` are still reported.

## What changed

- Removed `@gorhom/bottom-sheet` from the flagged package list.
- Removed the ambiguous `RN v7+` wording and narrowed the `formSheet` recommendation.
- Added regression tests for the Gorhom false positive and the remaining legacy-package warning.

## Test plan

```bash
pnpm --filter oxlint-plugin-react-doctor exec vp test run src/plugin/rules/react-native/rn-bottom-sheet-prefer-native.test.ts
```

### DIFF
--- a/packages/oxlint-plugin-react-doctor/src/plugin/rules/react-native/rn-bottom-sheet-prefer-native.test.ts
+++ b/packages/oxlint-plugin-react-doctor/src/plugin/rules/react-native/rn-bottom-sheet-prefer-native.test.ts
@@ -1,0 +1,22 @@
+import { describe, expect, it } from "vite-plus/test";
+import { runRule } from "../../../test-utils/run-rule.js";
+import { rnBottomSheetPreferNative } from "./rn-bottom-sheet-prefer-native.js";
+
+describe("rn-bottom-sheet-prefer-native", () => {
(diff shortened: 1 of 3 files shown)


aidenybai:
bugbot run

PR #481, opened 2026-05-25
[MIL-1550] fix navigation and Expo image false positives

rayhanadev:
## Why

Fixes two rule false positives:

- `react-compiler-destructure-method` should not ask React Navigation users to destructure `useNavigation()` methods. React Navigation returns a navigation object whose methods are meant to stay on that object.
- `rn-prefer-expo-image` should not ask bare React Native projects to add `expo-image`. `expo-image` is an Expo SDK package, so the recommendation only fits Expo or already `expo-image`-aware packages.

Before:

```tsx
import { useNavigation } from "@react-navigation/native";

const Screen = () => {
  const navigation = useNavigation();
  navigation.navigate("Chat", { sessionId });
};
```

```tsx
import { Image } from "react-native";

export const Screen = () => <Image source={{ uri }} />;
```

After: React Navigation method calls and bare React Native `Image` imports are not reported. Expo packages still get the `expo-image` recommendation, and non-React-Navigation hook objects still get the destructuring recommendation.

## What changed

- Skips `react-compiler-destructure-method` only for `useNavigation` imported from `@react-navigation/native` or `@react-navigation/core`.
- Keeps reporting same-named `useNavigation` hooks from unrelated packages or local code.
- Gates `rn-prefer-expo-image` on the nearest package declaring Expo or `expo-image`, with a fallback to the project-level Expo framework hint when package data is unavailable.
- Adds regression coverage for React Navigation native/core imports, unrelated `useNavigation`, bare RN `Image`, Expo `Image`, and mixed Expo + bare-RN workspaces.

## Test plan

```sh
pnpm --filter oxlint-plugin-react-doctor build && pnpm --filter @react-doctor/core build && pnpm exec vp test run packages/react-doctor/tests/regressions/architecture-rules.test.ts packages/react-doctor/tests/regressions/rn-package-scoping.test.ts
pnpm --filter @react-doctor/api build && pnpm --filter react-doctor typecheck
```

### DIFF
--- a/packages/oxlint-plugin-react-doctor/src/plugin/rules/architecture/react-compiler-destructure-method.ts
+++ b/packages/oxlint-plugin-react-doctor/src/plugin/rules/architecture/react-compiler-destructure-method.ts
@@ -18,11 +18,22 @@ const HOOK_OBJECTS_WITH_METHODS = new Map<string, Set<string>>([
   ["useSearchParams", new Set(["get", "getAll", "has", "set"])],
 ]);
 
-const REACT_NAVIGATION_MODULES = ["@react-navigation/native", "@react-navigation/core"];
+// Some libraries expose method-bearing hook objects where destructuring is not
+// part of the supported API shape, even though the hook name and method access
+// look like a normal React Compiler candidate. Keep those carve-outs keyed by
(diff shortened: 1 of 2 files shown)


aidenybai:
bugbot run

PR #484, opened 2026-05-25
[MIL-1554] Skip prefilled project question outside monorepos

rayhanadev:
## Why?

React Doctor's CLI was treating any single React project discovered under the scan root as a project selection result. For a normal non-monorepo app, that produced a prefilled "Select projects to scan" line even though there is nothing for the user to choose. This made the interactive flow look like it asked a project question outside monorepos, contradicting the intended behavior from MIL-1554. The fix keeps project selection as a monorepo-only concern and lets single-package apps proceed directly.

## What changed?

- Before: `selectProjects` fell back to `discoverReactSubprojects` whenever `listWorkspacePackages` returned no packages. For a non-monorepo React app, discovery returned the root app and printed a prefilled project selection line.
- After: `selectProjects` first checks `isMonorepoRoot` before falling back to subproject discovery. Non-monorepo roots now return `[rootDirectory]` silently.
- Added regression tests for both the non-monorepo single-app path and the one-workspace monorepo path, so the new skip behavior does not remove the monorepo selection output.

## Test plan

Test users can run to verify correctness:

- `cd packages/react-doctor && nr test tests/select-projects.test.ts`
- `cd packages/react-doctor && nr typecheck`
- `nr lint`
- `nr format:check`

---

> [!NOTE]
> **Low Risk**
> CLI UX-only change in project selection with targeted regression tests; no auth, data, or scan logic changes.
> 
> **Overview**
> **Non-monorepo apps** no longer show a prefilled *Select projects to scan* line. When `listWorkspacePackages` is empty, `selectProjects` now returns the scan root immediately if `isMonorepoRoot` is false, instead of always calling `discoverReactSubprojects` (which treated a lone React app as a one-package “selection”).
> 
> **Monorepos** keep the previous behavior: empty workspace lists still fall through to subproject discovery, and a single discovered package still logs the selection line.
> 
> New tests cover the silent non-monorepo path, a one-workspace monorepo (log still shown), and monorepo + nested React app via discovery.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5f6d31cbd1a4566af14f0e1a0860d8c57951fcf3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/react-doctor/src/cli/utils/select-projects.ts
+++ b/packages/react-doctor/src/cli/utils/select-projects.ts
@@ -1,7 +1,11 @@
 import path from "node:path";
 import type { WorkspacePackage } from "@react-doctor/core";
-import { discoverReactSubprojects, listWorkspacePackages } from "@react-doctor/core";
-import { highlighter } from "@react-doctor/core";
+import {
+  discoverReactSubprojects,
+  highlighter,
+  isMonorepoRoot,
+  listWorkspacePackages,
+} from "@react-doctor/core";
 import { cliLogger as logger } from "./cli-logger.js";
 import { prompts } from "./prompts.js";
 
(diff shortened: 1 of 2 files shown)


aidenybai:
bugbot run

PR #487, opened 2026-05-25
[MIL-1555] False positive for `tanstack-start-missing-head-content`

### LINKED ISSUE #479: False positive react-doctor/tanstack-start-missing-head-content
### Summary

After satisfying rule `react-doctor/only-export-components` and spliting exports for const and component running React Doctor triggers the `react-doctor/tanstack-start-missing-head-content` error.

It seems like a false positive because the root component contains the `<HeadContent />` which itself is included in the root route as the shell component. I might be wrong but the split seems legit and doesn't break anything. Correct me if I'm wrong, thanks.

### Diagnostics error

```
  ⚠ react-doctor/tanstack-start-missing-head-content
      Root route (__root) without <HeadContent /> — route head() meta tags won't render
      → Add `<HeadContent />` inside `<head>` in your __root route — without it, route `head()` meta tags are silently dropped
      src/routes/__root.tsx:1
```

### Source code

`src/components/root-document.tsx`:

```tsx
import { TanStackDevtools } from "@tanstack/react-devtools";
import { ReactQueryDevtoolsPanel } from "@tanstack/react-query-devtools";
import { HeadContent, Scripts } from "@tanstack/react-router";
import { TanStackRouterDevtoolsPanel } from "@tanstack/react-router-devtools";
import { ThemeProvider } from "@/components/providers/theme";

export function RootDocument({ children }: { children: React.ReactNode }) {
  return (
    <html lang="en" suppressHydrationWarning>
      <head>
        <HeadContent />
      </head>
      <body>
        <ThemeProvider defaultTheme="system" storageKey="theme">
          <div className="root">{children}</div>
        </ThemeProvider>
        <TanStackDevtools
          config={{ position: "bottom-right" }}
          plugins={[
            {
              name: "Tanstack Router",
              render: <TanStackRouterDevtoolsPanel />,
            },
            { name: "Tanstack Query", render: <ReactQueryDevtoolsPanel /> },
          ]}
        />
        <Scripts />
      </body>
    </html>
  );
}
```

`src/routes/__root.tsx`:

```tsx
import interLatinWoff2 from "@fontsource-variable/inter/files/inter-latin-wght-normal.woff2?url";
import type { QueryClient } from "@tanstack/react-query";
import { createRootRouteWithContext } from "@tanstack/react-router";
import { NotFound } from "@/components/not-found";
import { RootDocument } from "@/components/root-document";
import appCss from "@/styles.css?url";

export const Route = createRootRouteWithContext<{ queryClient: QueryClient }>()(
  {
    head: () => ({
      meta: [
        {
          charSet: "utf-8",
        },
        {
          name: "viewport",
          content: "width=device-width, initial-scale=1",
        },
        {
          title: "Redacted",
        },
        {
          name: "description",
          content: "Redacted.",
        },
      ],
      links: [
        {
          rel: "stylesheet",
          href: appCss,
        },
        {
          rel: "icon",
          href: "/favicon.svg",
          type: "image/svg+xml",
        },
        {
          rel: "preload",
          as: "font",
          type: "font/woff2",
          href: interLatinWoff2,
          crossOrigin: "anonymous",
        },
      ],
    }),
    notFoundComponent: NotFound,
    shellComponent: RootDocument,
  },
);

```

rayhanadev:
## Why?

TanStack Start apps can define their root document shell in ways that do not place a literal `<HeadContent />` element directly in the same simple shape the rule previously expected. The old `tanstack-start-missing-head-content` check reported every `__root` route file that did not contain a bare `HeadContent` JSX identifier, which created false positives for valid root routes that delegate the document shell elsewhere or render `HeadContent` through an alias or TanStack Router namespace import.

The rule should still catch the real bug: a root document that renders a `<head>` element but omits TanStack Router's `HeadContent`, causing route `head()` metadata to be dropped. This change narrows the diagnostic to that concrete document-head case and recognizes common valid import/render forms.

## What changed?

- Updated `tanstack-start-missing-head-content` to report only when a TanStack Start `__root` file renders a document `<head>` without `HeadContent`.
- Added import tracking for `HeadContent` aliases from `@tanstack/react-router`, so `<RouterHeadContent />` is treated as valid.
- Added namespace import tracking for `@tanstack/react-router`, so `<TanStackRouter.HeadContent />` is treated as valid.
- Added regression coverage for failing and passing cases.

Before:

```tsx
export const Route = createRootRoute({
  shellComponent: RootDocument,
})
```

This could be flagged even though the root route delegates document rendering.

After:

```tsx
export const Route = createRootRoute({
  shellComponent: RootDocument,
})
```

This is allowed because the file does not render a document `<head>` missing `HeadContent`.

Before:

```tsx
import { HeadContent as RouterHeadContent } from "@tanstack/react-router"

<head>
  <RouterHeadContent />
</head>
```

This could be flagged because only the literal `HeadContent` JSX name was recognized.

After:

```tsx
import { HeadContent as RouterHeadContent } from "@tanstack/react-router"

<head>
  <RouterHeadContent />
</head>
```

This is allowed because the alias resolves to TanStack Router's `HeadContent`.

## Test plan

Users can verify correctness by running:

```bash
pnpm --filter oxlint-plugin-react-doctor test src/plugin/rules/tanstack-start/tanstack-start-missing-head-content.test.ts
pnpm --filter react-doctor test tests/run-oxlint/tanstack-start.test.ts
pnpm format:check
```

---

> [!NOTE]
> **Low Risk**
> Lint-rule heuristic and test-harness changes only; no runtime app or security impact, with expanded tests limiting regression risk.
> 
> **Overview**
> Narrows **`tanstack-start-missing-head-content`** so it only warns when a TanStack Start **`__root`** file actually renders a document **`<head>`** that lacks **`HeadContent`** (or a plausible stand-in), instead of flagging every root route missing a literal `<HeadContent />` JSX name.
> 
> The rule now **pre-scans imports and aliases** (named/barrel **`HeadContent`**, **`@tanstack/react-router`** namespace **`HeadContent`**, and simple variable aliases) and treats **custom components under `<head>`** as safe. Routes that only use **`shellComponent`** or non-root files are no longer reported. **`runRule`** test harness now runs **`Program:exit`** visitors so exit-time diagnostics match oxlint.
> 
> Adds a broad regression test suite for pass/fail cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ffd96fc340c09407a4285088dedc372bb3859ba1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/oxlint-plugin-react-doctor/src/plugin/rules/tanstack-start/tanstack-start-missing-head-content.test.ts
+++ b/packages/oxlint-plugin-react-doctor/src/plugin/rules/tanstack-start/tanstack-start-missing-head-content.test.ts
@@ -1,0 +1,186 @@
+import { describe, expect, it } from "vite-plus/test";
+import { attachParentReferences } from "../../../test-utils/attach-parent-references.js";
+import { parseFixture } from "../../../test-utils/parse-fixture.js";
+import { analyzeControlFlow } from "../../semantic/control-flow-graph.js";
+import { analyzeScopes } from "../../semantic/scope-analysis.js";
+import type { EsTreeNode } from "../../utils/es-tree-node.js";
+import { isAstNode } from "../../utils/is-ast-node.js";
+import type { ReportDescriptor } from "../../utils/report-descriptor.js";
+import type { RuleContext } from "../../utils/rule-context.js";
+import type { RuleVisitors } from "../../utils/rule-visitors.js";
+import { tanstackStartMissingHeadContent } from "./tanstack-start-missing-head-content.js";
+
+const ROOT_ROUTE_FILENAME = "src/routes/__root.tsx";
+
+interface RuleDiagnostic {
+  message: string;
+  nodeType: string;
+}
+
+interface RunMissingHeadContentRuleResult {
+  diagnostics: RuleDiagnostic[];
+  parseErrors: ReadonlyArray<{ message: string }>;
+}
+
+const dispatchTreeWalkWithProgramExit = (root: EsTreeNode, visitors: RuleVisitors): void => {
+  const visit = (node: EsTreeNode): void => {
+    const handler = visitors[node.type];
+    if (typeof handler === "function") handler(node);
+
+    const nodeRecord = node as unknown as Record<string, unknown>;
+    for (const key of Object.keys(nodeRecord)) {
+      if (key === "parent") continue;
+      const child = nodeRecord[key];
+      if (Array.isArray(child)) {
+        for (const item of child) {
+          if (isAstNode(item)) visit(item);
+        }
+      } else if (isAstNode(child)) {
+        visit(child);
+      }
+    }
+  };
+
+  visit(root);
+  const programExitHandler = visitors["Program:exit"];
+  if (typeof programExitHandler === "function") programExitHandler(root);
+};
+
+const runMissingHeadContentRule = (
+  code: string,
+  filename = ROOT_ROUTE_FILENAME,
+): RunMissingHeadContentRuleResult => {
+  const parsed = parseFixture(code, { filename });
+  attachParentReferences(parsed.program);
+
+  const diagnostics: RuleDiagnostic[] = [];
+  const context: RuleContext = {
+    report: (descriptor: ReportDescriptor) => {
+      diagnostics.push({
+        message: descriptor.message,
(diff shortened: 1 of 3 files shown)


aidenybai:
bugbot run

PR #506, opened 2026-05-27
feat: animate perfect score terminal header

rayhanadev:
## Why?

A perfect `100/100` React Doctor score should feel meaningfully different in the terminal without changing scoring behavior or adding noise for non-perfect runs. This PR adds a terminal-only celebration for perfect scores while keeping the final resting output readable: the doctor face, score text, and branding return to the normal success styling, and the completed progress bar remains rainbow.

## What changed?

- Added a dedicated perfect-score rendering path in `render-score-header.ts`.
- Animates the full terminal score header during the `100/100` celebration:
  - doctor face
  - score text
  - progress bar
  - branding row
- Uses a smooth OKLCH-based true-color rainbow so hues shift without harsh brightness jumps.
- Aligns the rainbow across each full terminal row instead of restarting per text fragment.
- Settles the final output back to normal face/text/branding colors, with only the completed bar staying rainbow.

Before: `100/100` rendered like a normal green score header.

After: `100/100` animates the full score header briefly, then rests with normal text colors and a rainbow completed bar.

Perfect score terminal preview

## Test plan

Run these from the repo root:

```sh
pnpm --filter react-doctor typecheck
pnpm --filter react-doctor exec vp test run tests/render-score-header.test.ts
```


rayhanadev:
bugbot run

PR #546, opened 2026-05-28
fix(oxlint-plugin): skip js-tosorted-immutable on React Native/Expo (Hermes)

### LINKED ISSUE #543: Bug: false positive js-tosorted-immutable recommendation crashes Hermes (RN/Expo)
## Summary

The `js-tosorted-immutable` rule recommends `Array.prototype.toSorted()` as the immutable replacement for `[...arr].sort()`. In a React Native app running on Hermes (the default JS engine in current Expo SDKs), calling `.toSorted(...)` throws `TypeError: undefined is not a function` at runtime — Hermes hasn't shipped the ES2023 change-array-by-copy methods yet.

So the rule, applied via `--fix` or manually, can turn a working file into a crashing one.

## Reproduction

- Project: Expo SDK 56, React Native 0.85.3, Hermes enabled (default)
- TypeScript / `lib` setting: `expo/tsconfig.base` → `lib: ["esnext"]`, so TS happily type-checks `.toSorted(...)`
- react-doctor 0.2.6 / 0.2.9

Before:

```ts
const sorted = preferred
  ? [...cardType].sort((a, b) => (b.id === preferred ? 1 : 0) - (a.id === preferred ? 1 : 0))
  : cardType;
```

react-doctor warning:

```
⚠ react-doctor/js-tosorted-immutable
    [...array].sort() — use array.toSorted() for immutable sorting (ES2023)
    → Use `array.toSorted()` (ES2023) instead of `[...array].sort()` for immutable sorting without the spread allocation
    src/stores/useCardStore.ts:100
```

After applying the suggestion:

```ts
const sorted = preferred
  ? cardType.toSorted((a, b) => (b.id === preferred ? 1 : 0) - (a.id === preferred ? 1 : 0))
  : cardType;
```

Runtime crash on iOS dev build:

```
ERROR  [Error: Uncaught (in promise, id: 0) TypeError: undefined is not a function]
src/stores/useCardStore.ts:100  cardType.toSorted(...)
```

The same applies to `toReversed`, `toSpliced`, and `Array.prototype.with` — none are in Hermes today.

## Affected rules

- `react-doctor/js-tosorted-immutable` (confirmed)
- Likely any companion rules that suggest the other ES2023 change-array-by-copy methods.

## Suggestion

A few options that would help:

1. **Detect Hermes / React Native target** (e.g. via `package.json` having `react-native` as a dep, or an Expo config) and downgrade or skip these ES2023 rules.
2. **Make the rule configurable** with a `target` / `engines` option so projects can opt out when they ship to Hermes.
3. **Mention the Hermes caveat in the rule message** (e.g. "Note: not yet supported in Hermes — disable this rule for React Native projects on current Expo/RN SDKs.").

Option 1 would be the least surprising for RN users; option 2 gives an escape hatch when Hermes catches up.

## Tracking

Hermes ECMAScript compatibility: <https://github.com/facebook/hermes>. As of this report, the change-array-by-copy proposal isn't shipped.

## Environment

- react-doctor: 0.2.9
- Node: v22.12.0
- OS: macOS (darwin arm64)
- Expo SDK: 56
- React Native: 0.85.3
- JS Engine: Hermes

rayhanadev:
## Why

Catches a false positive that turns working React Native / Expo code into a runtime crash. Fixes #543.

`js-tosorted-immutable` is a `global` rule that recommends rewriting `[...array].sort()` as `array.toSorted()`. Hermes — the default JS engine in current Expo/RN SDKs — hasn't shipped the ES2023 change-array-by-copy methods, so applying the suggestion (manually or via `--fix`) throws `TypeError: undefined is not a function` at runtime. TypeScript happily type-checks it (`lib: ["esnext"]`), so nothing catches it before the device does.

Before (what the rule pushed RN users toward — crashes on Hermes):

```ts
const sorted = preferred
  ? cardType.toSorted((a, b) => (b.id === preferred ? 1 : 0) - (a.id === preferred ? 1 : 0))
  : cardType;
// ERROR  TypeError: undefined is not a function
```

After (rule no longer fires in RN/Expo, so the working spread-sort is left alone):

```ts
const sorted = preferred
  ? [...cardType].sort((a, b) => (b.id === preferred ? 1 : 0) - (a.id === preferred ? 1 : 0))
  : cardType;
```

## What changed

- Gated `js-tosorted-immutable` with `disabledBy: ["react-native"]`.
- The rule now only fires where the engine supports `toSorted()` (web). It no-ops in Expo, React Native, and monorepos that contain an RN workspace — reusing the `react-native` capability `buildCapabilities` already emits.
- Mirrors the existing `disabledBy: ["react-compiler"]` pattern on the `jsx-no-new-*-as-prop` rules — no new infrastructure and no codegen change (the registry imports the rule object, so `disabledBy` flows straight into `createOxlintConfig`'s gate).
- Scope check: the other `toSorted`/`toReversed`/`toSpliced`/`with` references in the codebase are detection sets (`FRESH_ARRAY_METHODS`, `DERIVING_ARRAY_METHODS`), not suggestions, so `js-tosorted-immutable` is the only rule that needed gating.
- Adds regression tests (flagged on web; suppressed on Expo and bare React Native) and a patch changeset.

## Eval results

RDE not run. This change can only narrow an existing rule — it removes diagnostics in RN/Expo projects and changes nothing on web — so it cannot introduce new false positives. The project-capability gate is deterministic and is covered by the end-to-end regression tests below.

## Test plan

- `pnpm exec vp test run tests/regressions/js-performance-rules.test.ts` — 60 passed (incl. 3 new issue #543 cases)
- `pnpm exec turbo run typecheck --filter=react-doctor --filter=oxlint-plugin-react-doctor` — pass
- `pnpm exec vp lint <changed files>` — pass
- `pnpm exec vp fmt --check <changed files>` — pass

---

> [!NOTE]
> **Low Risk**
> Narrows when one lint rule runs; web behavior unchanged and only removes unsafe suggestions on RN/Expo.
> 
> **Overview**
> Fixes a **false positive** where `js-tosorted-immutable` pushed React Native / Expo apps toward `array.toSorted()`, which **Hermes does not implement**, so fixes could crash at runtime while `[...array].sort()` still works.
> 
> The rule now sets **`disabledBy: ["react-native"]`**, reusing the existing project-capability gate (same pattern as other `disabledBy` rules). It **still runs on web**; it is **off for Expo, bare React Native, and RN workspaces**. Regression tests cover web vs Expo vs RN, plus a patch changeset.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0a41698ba649fe27b06ce1d1478401ee86c1c414. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/tosorted-immutable-hermes.md
+++ b/.changeset/tosorted-immutable-hermes.md
@@ -1,0 +1,5 @@
+---
+"oxlint-plugin-react-doctor": patch
+---
+
+Stop `js-tosorted-immutable` from firing in React Native / Expo projects. Hermes (the default RN/Expo JS engine) hasn't shipped the ES2023 change-array-by-copy methods, so the rule's recommended `array.toSorted()` rewrite of `[...array].sort()` crashed at runtime with `TypeError: undefined is not a function`. The rule now carries `disabledBy: ["react-native"]`, so it only fires on projects whose engine supports `toSorted()`.
--- a/packages/oxlint-plugin-react-doctor/src/plugin/rules/js-performance/js-tosorted-immutable.ts
+++ b/packages/oxlint-plugin-react-doctor/src/plugin/rules/js-performance/js-tosorted-immutable.ts
@@ -9,6 +9,12 @@ export const jsTosortedImmutable = defineRule<Rule>({
   id: "js-tosorted-immutable",
   tags: ["test-noise"],
   severity: "warn",
+  // Hermes (the default React Native / Expo JS engine) hasn't shipped
+  // the ES2023 change-array-by-copy methods, so `array.toSorted()`
+  // throws `undefined is not a function` at runtime. Recommending it in
+  // an RN/Expo project would turn working `[...array].sort()` code into
+  // a crash, so the gate drops this rule there. See issue #543.
+  disabledBy: ["react-native"],
   recommendation:
     "Use `array.toSorted()` (ES2023) instead of `[...array].sort()` for immutable sorting without the spread allocation",
   create: (context: RuleContext) => ({
--- a/packages/react-doctor/tests/regressions/js-performance-rules.test.ts
+++ b/packages/react-doctor/tests/regressions/js-performance-rules.test.ts
@@ -1363,3 +1363,56 @@ describe("async-parallel", () => {
     expect(hits).toHaveLength(0);
   });
 });
+
+describe("issue #543: js-tosorted-immutable is gated off for React Native / Expo (Hermes)", () => {
+  // Hermes — the default RN/Expo JS engine — hasn't shipped the ES2023
+  // change-array-by-copy methods, so `array.toSorted()` throws at
+  // runtime. Recommending it (and `--fix`-ing to it) turns working
+  // `[...array].sort()` code into a crash, so the rule must not fire in
+  // a React Native / Expo project. It stays on for web projects, where
+  // every modern engine supports `toSorted()`.
+  const SPREAD_SORT_SOURCE = `
+    export const sortCards = (
+      cardType: Array<{ id: string }>,
+      preferred: string | null,
+    ) =>
+      preferred
+        ? [...cardType].sort(
+            (first, second) =>
+              (second.id === preferred ? 1 : 0) - (first.id === preferred ? 1 : 0),
+          )
+        : cardType;
+  `;
+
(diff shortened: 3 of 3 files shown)


rayhanadev:
bugbot run

PR #548, opened 2026-05-28
feat(rn): hint Reanimated .get()/.set() on react-compiler immutability findings

### LINKED ISSUE #544: react-hooks-js/immutability on reanimated SharedValue: downgrade severity and surface the .get()/.set() alternative in the message
## Scope

**Ask:** don't mark `react-hooks-js/immutability` as ERROR when it fires on a `react-native-reanimated` `SharedValue`, and surface the Compiler-compatible alternative (`sharedValue.get()` / `sharedValue.set(...)`) directly in the rule message.

> Updated after a closer read of the reanimated docs — the original framing of this as a flat false positive was wrong. The rule is genuinely catching something the user can act on; the gap is in *how* it's reported. See the [follow-up comment](#issuecomment-4566592298) and the rewritten body below.

## The actual problem

`react-hooks-js/immutability` fires on the common reanimated pattern:

```ts
const rotation = useSharedValue(0);
useEffect(() => {
  rotation.value = withRepeat(withTiming(360, { duration }), -1, false);
}, [duration, rotation]);

const style = useAnimatedStyle(() => ({
  transform: [{ rotate: `${rotation.value}deg` }],
}));
```

react-doctor reports:

```
✗ react-hooks-js/immutability
    React Compiler can't optimize this code
    → Error: This value cannot be modified

    Modifying a value returned from a hook is not allowed. Consider moving the
    modification into the hook where the value is constructed.
    src/components/SpinningIcon.tsx:23
```

The finding is real — React Compiler treats hook returns as immutable and the `.value =` mutation breaks that contract, so Compiler skips the component. Reanimated documents this and ships an alternative: `SharedValue.get()` / `SharedValue.set(...)` (3.16+) are the Compiler-friendly accessors. The fix is mechanical:

```ts
rotation.set(withRepeat(withTiming(360, { duration }), -1, false));
const style = useAnimatedStyle(() => ({ transform: [{ rotate: `${rotation.get()}deg` }] }));
```

## What's wrong with the current report

1. **Severity:** marked as ERROR, which implies "must fix and the code is broken." The Compiler skipping one component because it delegates to reanimated's UI-thread runtime isn't broken — it's a known boundary. A WARN with context fits better.
2. **No actionable hint:** the message says "Consider moving the modification into the hook where the value is constructed." That's the generic React Compiler suggestion and isn't applicable to library-owned values like `SharedValue`. There's no mention of `.get()` / `.set()`, the actual fix.
3. **Result:** users either chase the rule into worse designs (rewrites away from reanimated), file upstream issues like the original framing of this one, or learn to ignore "React Compiler can't optimize" entirely — which is bad because *sometimes* it's a real signal on regular code.

## Suggested fix (ranked by ambition)

1. **Detect reanimated `useSharedValue` source and customize the message:** when the mutated identifier originates from `react-native-reanimated`'s `useSharedValue`, downgrade severity to WARN and append something like _"reanimated 3.16+ ships `.get()` / `.set()` accessors for React Compiler compatibility — prefer those over `.value` here."_
2. **Generalize to a "library boundary" allowlist:** project config can register libraries that own their own UI-thread perf (`react-native-reanimated`, `@shopify/react-native-skia`, …); for findings sourced from their hook returns, downgrade severity and link to a per-library hint.
3. **Documentation-only:** rule message acknowledges that some libraries (reanimated specifically) have Compiler-friendly accessors and links to their docs. Cheaper but still meaningfully better than the current generic message.

Even just (3) would have saved a round trip here.

## Environment

- react-doctor: 0.2.9
- React: 19.2.3
- React Compiler: enabled (Expo SDK 56 default)
- react-native-reanimated: 4.3.1
- Expo SDK: 56
- React Native: 0.85.3
- Platform: iOS / Android (rule fires regardless)

## Related

- Reanimated docs on React Compiler compatibility: https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/shared-values
- Sibling issue with same "library can't satisfy what the rule asks for" shape but no good alternative: #543 (`js-tosorted-immutable` recommending Hermes-incompatible API)

rayhanadev:
## Why

`react-hooks-js/immutability` — the upstream React Compiler rule — fires on the canonical Reanimated `sharedValue.value = ...` pattern, because React Compiler treats values returned from hooks as immutable and skips optimizing the component. The rule's generic advice ("move the modification into the hook where the value is constructed") is impossible to follow for a library-owned `SharedValue`, so React Native users hit a dead end (and sometimes file upstream issues — see #544).

Reanimated ships `.get()` / `.set()` accessors specifically as the React Compiler-compatible API. react-doctor now surfaces that alternative directly in the diagnostic's help text.

Before:

```tsx
const rotation = useSharedValue(0);
useEffect(() => {
  // react-hooks-js/immutability ERROR — generic, non-actionable hint
  rotation.value = withRepeat(withTiming(360), -1);
}, [rotation]);
```

After:

```tsx
const rotation = useSharedValue(0);
useEffect(() => {
  // React Compiler-compatible accessor; clears the finding
  rotation.set(withRepeat(withTiming(360), -1));
}, [rotation]);
```

The diagnostic help now ends with an actionable pointer:

```text
React Compiler can't optimize this code
→ Modifying a value returned from a hook is not allowed. ...

  If this is a Reanimated shared value, prefer its React Compiler-compatible
  `.get()` / `.set()` accessors over `.value` —
  https://docs.swmansion.com/react-native-reanimated/docs/core/useSharedValue/#react-compiler-support
```

## What changed

- Added `appendReanimatedSharedValueHint` util in `@react-doctor/core` and wired it into the `react-hooks-js` branch of `parseOxlintOutput`.
- Appends the Reanimated `.get()` / `.set()` hint to the help text of the `immutability` diagnostic only.
- Gated to `rule === "immutability"` **and** React Native projects (`project.hasReactNativeWorkspace`) — the other ~15 React Compiler rules and web/Next/Vite scans are untouched.
- Phrased conditionally ("If this is a Reanimated shared value…") so it stays accurate even when an immutability finding isn't a `SharedValue`.
- **Severity deliberately unchanged.** The reporter's original "false positive / downgrade to WARN" framing was retracted in their follow-up: the finding is genuinely actionable (Compiler does skip the component), and the parser has no AST to tell a `SharedValue` mutation apart from any other illegal hook-return mutation, so a blanket downgrade would silence real errors.
- Adds unit + end-to-end (`parseOxlintOutput`) tests for RN/web gating and rule gating.

## Eval results

RDE not run — this is a deterministic help-text transformation on an existing upstream diagnostic, not a new AST/heuristic rule, so there is no detection surface to measure for noise. Behavior is covered by the new unit + parser integration tests instead.

## Test plan

- `pnpm --filter @react-doctor/core exec vp test run append-reanimated-shared-value-hint` — 7 passed
- `pnpm test` — full suite green (core 353, react-doctor 1378 passed / 12 skipped)
- `pnpm typecheck` — 8/8
- `pnpm lint` + `pnpm format:check` — clean

> Note: this branch also carries a pre-existing commit (`9a05ef20` "fix: copy should say scanning files", a 1-line `run-inspect.ts` change) that isn't on `main` yet and is unrelated to this change. It will drop out of this diff automatically if it lands via another PR first.

Closes #544

---

<a href="https://app.devin.ai/review/millionco/react-doctor/pull/548" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

---

> [!NOTE]
> **Low Risk**
> Changes are limited to project metadata detection and optional diagnostic help text; no runtime behavior, auth, or data handling.
> 
> **Overview**
> Adds **`hasReanimated`** to project discovery (root or any workspace declares `react-native-reanimated`, only after the project is treated as React Native) and refactors workspace walks through shared **`someWorkspacePackageJson`**, so RN and Reanimated detection use the same monorepo traversal.
> 
> For **`react-hooks-js` / `immutability`** React Compiler diagnostics, **`parseOxlintOutput`** now appends conditional help pointing at Reanimated’s **`.get()` / `.set()`** accessors when **`hasReanimated`** is true, without changing severity or other Compiler rules. **`ProjectInfo`** and test fixtures gain the new flag; unit and parser tests cover gating and end-to-end help text.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0816f8558ce6a02bac408488e1c418a9fd6f4ebf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/project-info/discover-project.ts
+++ b/packages/core/src/project-info/discover-project.ts
@@ -13,6 +13,8 @@ import { getDependencyDeclaration } from "./utils/get-dependency-declaration.js"
 import { hasReactNativeWorkspaceAnywhere } from "./has-react-native-workspace-anywhere.js";
 import { hasPreact } from "./has-preact.js";
 import { hasTanStackQuery } from "./has-tanstack-query.js";
+import { someWorkspacePackageJson } from "./some-workspace-package-json.js";
+import { isPackageJsonReanimatedAware } from "./utils/is-package-json-reanimated-aware.js";
 import { readPackageJson } from "./read-package-json.js";
 import { isCatalogReference, resolveCatalogVersion } from "./resolve-catalog-version.js";
 import { resolveEffectiveReactMajor } from "./resolve-effective-react-major.js";
@@ -157,6 +159,13 @@ export const discoverProject = (directory: string): ProjectInfo => {
     framework === "react-native" ||
     hasReactNativeWorkspaceAnywhere(directory, packageJson);
 
+  // Only walk for reanimated once we already know it's an RN project —
+  // reanimated implies React Native, so a web project can never declare
+  // it, and this skips the workspace walk entirely for web monorepos.
+  const hasReanimated =
+    hasReactNativeWorkspace &&
+    someWorkspacePackageJson(directory, packageJson, isPackageJsonReanimatedAware);
+
   const projectInfo: ProjectInfo = {
     rootDirectory: directory,
     projectName,
@@ -169,6 +178,7 @@ export const discoverProject = (directory: string): ProjectInfo => {
     hasTanStackQuery: hasTanStackQuery(packageJson),
     hasPreact: hasPreact(packageJson),
     hasReactNativeWorkspace,
+    hasReanimated,
     sourceFileCount,
   };
   cachedProjectInfos.set(directory, projectInfo);
--- a/packages/core/src/project-info/has-react-native-workspace-anywhere.ts
+++ b/packages/core/src/project-info/has-react-native-workspace-anywhere.ts
@@ -1,8 +1,5 @@
-import path from "node:path";
 import type { PackageJson } from "../types/index.js";
-import { getWorkspacePatterns } from "./get-workspace-patterns.js";
-import { readPackageJson } from "./read-package-json.js";
-import { resolveWorkspaceDirectories } from "./resolve-workspace-directories.js";
+import { someWorkspacePackageJson } from "./some-workspace-package-json.js";
 import { isPackageJsonReactNativeAware } from "./utils/is-package-json-react-native-aware.js";
 
 // True when the root manifest or any workspace package inside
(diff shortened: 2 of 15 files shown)


rayhanadev:
bugbot run

rayhanadev:
bugbot run

PR #550, opened 2026-05-28
fix(ci): stabilize Windows vitest worker pool

rayhanadev:
## Summary

`test (windows-latest, 22.18.0)` was failing even though **0 test assertions fail**. Root cause: PR #537 moved the deslop dead-code scan into a native-addon `worker_threads.Worker`, and tearing that worker down inside a vitest fork intermittently crashes the host process **on Windows only**. This PR runs the dead-code analysis in a **child process** instead (clean to kill on every platform), which fixes the crash at its source for CI and for real Windows users — no skipped tests, no weakened assertions, no retries.

## Symptom

```
Tests   1333 passed | 60 skipped (1397)
Errors  2 errors
Error: [vitest-pool]: Worker forks emitted error.
  ❯ ChildProcess.emitUnexpectedExit .../@voidzero-dev/vite-plus-test/dist/chunks/cli-api.*.js
Caused by: Error: Worker exited unexpectedly
```

A forked vitest worker exits unexpectedly at teardown (the passed-count varies because the tests in-flight in the dying worker never report; GitHub logs "Cleaning up orphan processes"). Intermittent, Windows-only.

## Root cause — PR #537 (`fix(cli): avoid dead-code scan freezes`)

The Windows job was green **16 runs in a row** right up to #537, then failed on (nearly) every run after. #537 made two changes:

1. **The real cause.** It moved the deslop dead-code scan from an inline `await analyze(...)` into a **terminable `worker_threads.Worker`** (a legitimate fix so a hung scan can be killed via a 120s timeout). But that worker loads native (oxc) NAPI addons, and **creating + `terminate()`-ing native-addon worker threads inside a vitest forked test process is intermittently fatal on Windows** — the host fork dies silently. It scales with spawn count: `inspect()`/`diagnose()` default `deadCode: true`, so react-doctor ran it ~25×/run (failed almost always), and core's own `check-dead-code.test.ts` (~2×) failed intermittently too.
2. **A failed mitigation.** It also added `singleFork` / `maxWorkers: 1` / `fileParallelism: false` (win32 only) to `packages/react-doctor/vite.config.ts`, which just funnels every file into one long-lived worker and concentrates the same crash.

Pre-#537, deslop ran **inline** (no worker thread), so the identical heavy oxlint + oxc-native workload was green — confirming the worker-thread teardown, not general native load, is the trigger.

## Changes

| File | Change |
| --- | --- |
| `packages/core/src/check-dead-code.ts` | **Root-cause fix:** run deslop in a child PROCESS (`node -e`, input via stdin, result as JSON on stdout) instead of a `worker_threads.Worker`. Killed cleanly via SIGKILL — same pattern as the oxlint runner — so teardown never crashes the host fork. The `DeadCodeWorkerFactory` seam, 120s timeout, and result parsing are unchanged → identical behavior/output. |
| `packages/react-doctor/vite.config.ts` | Drop #537's win32 `singleFork` override; react-doctor runs on the default parallel + isolated forks (the green-16/16 config). Linux/macOS unchanged. |
| `packages/react-doctor/tests/{inspect,diagnose,inspect-surface-filter}.test.ts`, `tests/regressions/cli-and-output.test.ts` | Pass `deadCode: false` in the pipeline tests that assert on lint / project resolution (not dead-code), so they don't pay for the deslop analysis they never check. Matches the `api` package convention; keeps every platform fast. |
| `packages/react-doctor/tests/dead-code-integration.test.ts` | **New.** Runs the real deslop worker through the public `diagnose()` API and *asserts* a `unused-file` / `plugin: "deslop"` diagnostic surfaces — net **more** dead-code coverage than before (a real assertion where there were none). |

Dead-code stays fully exercised on all platforms (core's `check-dead-code.test.ts` + the new integration test). No `continue-on-error`, no retries, no matrix entries dropped, no tests skipped/deleted.

## Evidence

Final commit `6dae8b0b` — [CI run](https://github.com/millionco/react-doctor/actions/runs/26603358307):

- **`test (windows-latest, 22.18.0)` green across 3 consecutive runs** (initial + 2 Windows-job re-runs), each with the full count and **0 unhandled worker errors**:
  `Test Files 106 passed | 4 skipped (110)`, `Tests 1338 passed | 60 skipped (1398)` — vs the crashing `1325–1333 passed … N errors`.
- The previously-crashing real-worker tests now pass on Windows: `check-dead-code.test.ts (6 tests) ✓`, `dead-code-integration.test.ts (1 test) ✓`.
- All other matrix entries green: ubuntu 20.19/22.18/24/25/26, macOS; plus lint, typecheck, format, CodeQL, and the built-CLI / JSON-report smoke checks.
- Linux/macOS react-doctor test-CPU actually dropped (~406s → ~302s).

## Test plan

- [x] `pnpm format:check`, `pnpm lint`, `pnpm typecheck` — clean
- [x] `pnpm test` locally — core real-worker dead-code tests + react-doctor integration test pass against the child-process worker
- [x] `test (windows-latest, 22.18.0)` green with full count, 0 worker errors, real dead-code worker tests running
- [x] 3 consecutive green Windows runs on the final commit
- [x] No other matrix job regresses; lint/typecheck/format/smoke stay green

---

> [!NOTE]
> **Medium Risk**
> Changes process isolation for deslop/oxc native teardown (child process + SIGKILL vs worker terminate), which affects all platforms' dead-code path but preserves the same public API, timeout, and diagnostic output shape.
> 
> **Overview**
> Fixes intermittent **Windows CI** failures where vitest forks died with **"Worker exited unexpectedly"** even when assertions passed, by changing how deslop dead-code analysis runs in isolation.
> 
> **`check-dead-code.ts`** no longer uses a **`worker_threads.Worker`** with native oxc NAPI inside the host. It spawns a **`node -e` child process** that reads config on **stdin** and returns normalized JSON on **stdout**, with **`SIGKILL`** on timeout/teardown (same idea as the oxlint runner). The factory seam, 120s timeout, and diagnostic mapping stay the same.
> 
> **Test suite:** removes the win32-only **single-fork / `maxWorkers: 1`** vitest override so Windows uses the default parallel fork pool again. Pipeline tests that only care about lint/project resolution pass **`deadCode: false`** to avoid redundant deslop runs; a new **`dead-code-integration.test.ts`** runs real deslop through **`diagnose()`** and asserts an **`unused-file`** diagnostic.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6dae8b0b62355208cf18179e31b2d22cb773fe00. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/check-dead-code.ts
+++ b/packages/core/src/check-dead-code.ts
@@ -1,6 +1,6 @@
+import { spawn } from "node:child_process";
 import fs from "node:fs";
 import path from "node:path";
-import { Worker } from "node:worker_threads";
 import type { Diagnostic, ReactDoctorConfig } from "./types/index.js";
 import { collectIgnorePatterns } from "./collect-ignore-patterns.js";
 import { DEAD_CODE_WORKER_TIMEOUT_MS, MILLISECONDS_PER_SECOND } from "./constants.js";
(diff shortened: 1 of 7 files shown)


rayhanadev:
bugbot run

rayhanadev:
bugbot run

PR #460, opened 2026-05-23
proposal: no-prop-types

NisargIO:
# Proposal: `react-doctor/no-prop-types`

> **Status**: 🟡 Auto-discovered draft proposal. **Not yet implemented.** Maintainer review wanted before any code lands.

|                        |                      |
| ---------------------- | -------------------- |
| Category               | `architecture`       |
| Severity               | `warn`               |
| Source cluster         | `NEW::no-prop-types` |
| Backing evidence units | 1                    |

## Why the bug exists

> The developer assumed React would continue executing `Component.propTypes` validators at runtime. In React 19 those checks are removed and silently ignored, so invalid props no longer trigger the intended validation path.

## Generality check

> Many React apps and component libraries historically used `propTypes` for runtime validation, especially JavaScript code and published UI packages. Any unrelated React 19 codebase relying on those checks can experience the same silent validation loss.

## Sources

Discovered by the [react-doctor-evals discovery flywheel](https://github.com/millionco/react-doctor-evals/pull/11) mining bug-fix evidence across React OSS repos. This proposal was sourced from the **accepted-review** signal (a reviewer commented + the author edited the file in response + the thread was marked resolved). Pipeline:

```
OSS repo -> GitHub GraphQL (PR review threads) -> acceptance filter (file changed between comment SHA and merge SHA) -> EvidenceUnit -> DraftAgent (gpt-5.5, xhigh reasoning) -> RuleDedupe -> THIS PR
```

### Backing evidence

- [`react-native-elements/react-native-elements` - `packages/base/src/AirbnbRating/SwipeRating.tsx` (AcceptedReviewFixMeta)](https://github.com/react-native-elements/react-native-elements/pull/3990#discussion_r2486066335) — reviewer: human @deepktp · PR #3990 · signal: resolved-and-changed
  > https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-proptypes-and-defaultprops

## Validation prompt

FP-aware guidance for the [react-review agent](https://github.com/millionco/react-review) when triaging this rule:

> Verify the reported `propTypes` is being used as React runtime prop validation in a React 19+ component. False positive: `propTypes` is only metadata consumed by a documentation or design-system tool and React validation is not expected. False positive: the uppercase receiver is not a React component but a custom class or namespace object with an unrelated `propTypes` property.

## Fix prompt

[...]

PR #570, opened 2026-05-29
feat(oxlint-plugin): add no-prop-types rule (React 19)

aidenybai:
## Why

Implements the [`no-prop-types` proposal (#460)](https://github.com/millionco/react-doctor/pull/460).

React 19 removed runtime `propTypes` validation **entirely** — React no longer reads `Component.propTypes`. Invalid props that previously logged a `console.error` from the propTypes validators now pass through silently, so a class of input-validation bugs becomes invisible after upgrading. The rule nudges teams to express the prop contract in TypeScript and move any genuinely runtime validation into explicit checks.

Before:

```tsx
import PropTypes from "prop-types";

export function Rating({ value }) {
  return <span>{value}</span>;
}

Rating.propTypes = {
  value: PropTypes.number, // React 19 never runs this — invalid props pass silently
};
```

After:

```tsx
interface Props {
  value?: number;
}

export function Rating({ value = 0 }: Props) {
  return <span>{value}</span>;
}
```

## What changed

- Added the `no-prop-types` rule under `architecture/` (`framework: "global"`, `severity: "warn"`).
- **Version check:** gated with `requires: ["react:19"]`. The capability builder only adds `react:19` when the detected React major is ≥ 19, so React 17/18 projects (where `propTypes` still runs) and projects with an **unknown** React version stay completely quiet. This is the core safeguard the proposal called out.
- Detects two idioms:
  - `Component.propTypes = …` assignments (including computed `Component["propTypes"] = …`).
  - `static propTypes` class fields, on both named classes and class expressions bound to an uppercase `const`.
- Reports only when the receiver is **component-cased** (uppercase first letter), mirroring the sibling `no-default-props` heuristic — `config.propTypes = …` and `class validator { static propTypes }` are not flagged.
- Allows TypeScript prop types, destructuring defaults, instance (non-static) `propTypes` fields, and any lowercase/non-component receiver.
- Registered in the generated rule registry; the ESLint mirror and presets pick it up automatically.
- Added a changeset.

### Note on tags

The auto-generated proposal suggested `["test-noise", "migration-hint"]`. I shipped it with `["test-noise"]` only, matching the closest sibling `no-default-props`. `migration-hint` (fire in test files too) is reserved for **import-based** deprecated-API rules where test files are a primary migration surface (`react-dom/test-utils`); `propTypes`, like `defaultProps`, is a **pattern-based** rule where flagging test fixtures would just add noise. This keeps false positives down, which was the explicit goal.

## False-positive considerations

- Pre-19 / unknown React version → suppressed by the `react:19` capability gate.
- Lowercase / non-component receivers (`config.propTypes`, `class validator`) → not flagged.
- `this.propTypes`, member-chain receivers (`Foo.Bar.propTypes`), compound assignments (`+=`), and dynamic computed keys (`Foo[name]`) → not flagged (conservative).
- Test/story/playground files → skipped via the `test-noise` wrapper.

## Test plan

- `vp test run packages/react-doctor/tests/regressions/react-19-migration-rules.test.ts` — added 7 detector cases (assignment, static class field, class expression, computed key, + 3 negatives) and 3 version-gating cases (18 → none, 19 → flagged, unknown → none). 43/43 pass.
- `pnpm test` — full filtered suite: 1400 passed / 15 skipped.
- `pnpm typecheck`, `pnpm lint`, `pnpm format:check`, `pnpm smoke:json-report` — all pass.

<div><a href="https://cursor.com/agents/bc-32fcb156-b99d-45ac-ac27-8e01acf97cdf"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-32fcb156-b99d-45ac-ac27-8e01acf97cdf"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.changeset/no-prop-types.md
+++ b/.changeset/no-prop-types.md
@@ -1,0 +1,7 @@
+---
+"oxlint-plugin-react-doctor": patch
+"eslint-plugin-react-doctor": patch
+"react-doctor": patch
+---
+
+Add the `no-prop-types` architecture rule. React 19 removed runtime `propTypes` validation entirely — React no longer reads `Component.propTypes`, so invalid props that used to log a console warning now pass silently. The rule flags `Component.propTypes = { ... }` assignments and `static propTypes` class fields on component-cased identifiers, and is version-gated to React 19+ (`requires: ["react:19"]`) so projects where `propTypes` still runs stay quiet. It steers users toward TypeScript prop types plus explicit runtime validation. See #460.
--- a/packages/oxlint-plugin-react-doctor/src/plugin/rule-registry.ts
+++ b/packages/oxlint-plugin-react-doctor/src/plugin/rule-registry.ts
@@ -183,6 +183,7 @@ import { noPermanentWillChange } from "./rules/performance/no-permanent-will-cha
 import { noPolymorphicChildren } from "./rules/correctness/no-polymorphic-children.js";
 import { noPreventDefault } from "./rules/correctness/no-prevent-default.js";
 import { noPropCallbackInEffect } from "./rules/state-and-effects/no-prop-callback-in-effect.js";
+import { noPropTypes } from "./rules/architecture/no-prop-types.js";
 import { noPureBlackBackground } from "./rules/design/no-pure-black-background.js";
 import { noReactChildren } from "./rules/react-builtins/no-react-children.js";
 import { noReactDomDeprecatedApis } from "./rules/architecture/no-react-dom-deprecated-apis.js";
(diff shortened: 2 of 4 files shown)


rayhanadev:
what in the unc

PR #575, opened 2026-05-29
feat(oxlint-plugin): add no-focus-on-mount rule

aidenybai:
Implements the `react-doctor/no-focus-on-mount` proposal from #315.

## What

Adds a `state-and-effects` rule that flags `focus()` calls which run **on mount** from an empty-deps `useEffect` / `useLayoutEffect`. Auto-running focus on mount can steal focus before the surrounding UI is ready and surprise users who never interacted with the field.

## Detector

Reports when a hook call is an effect hook with an **empty** dependency array (`[]`) and a non-computed `.focus()` member call is reached **during mount execution** of the effect callback.

"During mount execution" walks the effect body and descends into a nested function only when it actually runs as a consequence of mounting:

- the effect body itself,
- immediately-invoked functions (IIFEs), including parenthesized / TS-wrapped forms,
- callbacks handed to timers/schedulers (`setTimeout`, `setInterval`, `requestAnimationFrame`, `requestIdleCallback`, `queueMicrotask` — reusing the shared `TIMER_AND_SCHEDULER_DIRECT_CALLEE_NAMES` constant).

### Stays quiet on (FP guards / v1 non-goals)

- **Test, story, and playground files** — the rule is tagged `test-noise` (like `no-prop-callback-in-effect`, `prefer-use-effect-event`, `no-effect-event-handler`), since autofocus-on-mount is routine in non-production code.
- **Focus inside an event-listener / handler callback** registered on mount (`addEventListener`, stored handlers) — runs on user interaction, the headline false-positive from the proposal's validation prompt.
- **Focus restored in the effect cleanup** (`return () => …`) — runs on unmount.
- **Scheduler nested inside an event handler** — outer handler is pruned before the timer is reached.
- Effects with a **non-empty** dependency array (e.g. `[isOpen]`) — the proposal's negative fixture.
- Effects with **no** dependency array.
- Computed `["focus"]` access (method not statically provable).
- Non-effect hooks such as `useMemo` with empty deps.

## Notable edge cases handled

- `oxc-parser` preserves `ParenthesizedExpression` wrappers, so an IIFE's `callee` is the paren wrapper, not the arrow. `ascendThroughExpressionWrappers` climbs structurally — every paren/TS/optional-chain wrapper exposes its inner node as `.expression` — so the function's real syntactic position is compared against the enclosing call without enumerating wrapper types.
- The focus-call match runs through the canonical `stripParenExpression`, so parenthesized/TS-wrapped callees like `(el.focus)()` are matched too.

## Files

- `…/rules/state-and-effects/no-focus-on-mount.ts` — the rule.
- `…/rules/state-and-effects/no-focus-on-mount.test.ts` — 17 adversarial pass/fail cases.
- `…/rule-registry.ts` — regenerated via `pnpm gen` (auto-registers as `framework: global`, flows into the recommended config and the ESLint mirror automatically).
- `.changeset/no-focus-on-mount.md` — patch changeset for the three published packages.

## Validation

- `no-focus-on-mount.test.ts`: 17/17 passing.
- `pnpm typecheck` (oxlint plugin): passing.
- `vp lint` + `vp fmt --check` on changed files: passing.
- `gen:check` idempotent (no registry drift).
- Full CI on the prior revision (tests across Linux/macOS/Windows on Node 20–26, lint, typecheck, CodeQL, Bugbot) was green.

<div><a href="https://cursor.com/agents/bc-ce2f7cd3-e860-45ad-81cf-9139a2039f01"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-ce2f7cd3-e860-45ad-81cf-9139a2039f01"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


aidenybai:
Note this is bad, too fp (52/66 FPs on OSS repos)

PR #596, opened 2026-05-30
feat: new onboarding

### LINKED ISSUE #541: Feature request: Add CLI flag to skip "Copy issues to clipboard?" prompt in React Native Doctor
It would be useful to have a CLI flag to automatically skip the interactive prompt:

```bash
× Copy issues to clipboard? ..
```

Currently, this prompt appears every time `react-native doctor` is executed, which slows down automated workflows and repeated usage.

## Proposed solution

Add a flag such as:

```bash
react-native doctor --no-clipboard
```

that disables clipboard-related prompts entirely.

## Why this matters

* Better DX for frequent usage
* Cleaner CI/CD integration
* Easier scripting and automation
* Avoids repetitive manual interaction

## Additional context

The current behavior is fine for first-time users, but becomes annoying when running the command often during development/debugging sessions.

aidenybai:
CleanShot 2026-05-31 at 00 27 50@2x

---

> [!NOTE]
> **Medium Risk**
> Broad behavior change: default hidden warnings, renamed category keys for config overrides, and altered scan file sets affect CI and programmatic consumers; core scan paths are well-tested in diff.
> 
> **Overview**
> This PR reshapes how **react-doctor** scans, filters, and presents findings—not onboarding UI despite the PR title.
> 
> **Output & taxonomy:** Diagnostics roll into five buckets (**Security**, **Bugs**, **Performance**, **Accessibility**, **Maintainability**), with codegen collapsing old labels (e.g. Correctness → Bugs, Dead Code → Maintainability). Rules gain optional **`title`** fields and clearer messages (especially a11y/security). The post-scan view leads with a **“Top errors you should fix”** block (category + title, inline code frames with length/merge guards). **Ranking** for that block uses only the score API’s per-rule priority when available; otherwise scan order (categories alphabetically).
> 
> **Defaults & pipeline:** **`warning`** findings are **hidden by default** (`--warnings` / `"warnings": true` to show). The diagnostic pipeline drops warnings unless explicitly opted in via per-rule/category `"warn"`. **Dead-code** analysis is **skipped** when warnings are off (unless overrides would surface deslop findings), and the deslop child runs with **8GB heap** to avoid OOM on type-heavy repos.
> 
> **File scoping:** Scans skip generated bundles (`*.iife.js`, `*.umd.js`, etc.) and **content-sniff** large minified `.js` so counts and lint sets stay aligned; oxlint parsing applies the same filters in diff mode.
> 
> **Monorepo / inspect:** Multi-project runs can **suppress per-project “Scanned N files”** and report **unique** file paths plus combined timing via new `InspectOutput` fields. Config validation **warns** on stale `categories` / surface category keys instead of silently ignoring them.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fc9ed91be455da38591c4d2b5783df311b541273. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---

<a href="https://app.devin.ai/review/millionco/react-doctor/pull/596" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>


packages/core/src/run-inspect.ts
```diff
@@ -1,3 +1,4 @@
+import path from "node:path";
```
aidenybai, line 1: should we use the Effect path thing here? node service?

packages/core/src/project-info/constants.ts
```diff
@@ -1,5 +1,40 @@
 export const SOURCE_FILE_PATTERN = /\.(tsx?|jsx?)$/;
 
+// Bundler output — IIFE / UMD-global builds (e.g. tsup/rollup emitting
+// `widget.iife.js` or `sdk.global.js`, often dropped into `public/`
+// alongside real source rather than an ignored `dist/`) — is generated,
+// usually minified, and not worth linting. Flagging a 17k-line
+// `*.iife.js` bundle is pure noise, so these are excluded from every
+// source-file scan by default. Only `.js` is matched: these formats are
+// browser-global bundles, and `.cjs`/`.mjs` aren't source files anyway.
+export const GENERATED_BUNDLE_FILE_PATTERN = /\.(iife|global)\.js$/i;
```
aidenybai: think of other common patterns here

packages/core/src/runners/oxlint/parse-output.ts
```diff
@@ -1,11 +1,14 @@
+import path from "node:path";
```
aidenybai, line 1: effect service for node or no

packages/core/src/utils/is-large-minified-file.ts
```diff
@@ -0,0 +1,20 @@
...
+// minified. The size gate keeps whole-tree discovery from reading every
+// small source file just to check. Both `listSourceFiles` (the scanned set)
+// and `countSourceFiles` (the reported `sourceFileCount`) route through here
+// so the two can never diverge. Returns false on any stat error so an
+// unreadable file is kept / counted as usual.
+export const isLargeMinifiedFile = (absolutePath: string): boolean => {
+  let sizeBytes: number;
+  try {
+    sizeBytes = fs.statSync(absolutePath).size;
+  } catch {
```
aidenybai, line 15: this is a terribl terrible heuritic. maybe just check the number of characters on one line. it oculd also be mutli line but little code lines, find if there is one code line that is suepr long. think of a better heuristic, this is way to leaky

aidenybai: oh wait you laready do this. this hould not be a separate util right or am i trolling

packages/core/src/validate-config-types.ts
```diff
@@ -99,11 +124,13 @@ const validateSurfaceControls = (
   const validated: SurfaceControls = {};
```
aidenybai: make this var name more descriptive

packages/oxlint-plugin-react-doctor/scripts/generate-rule-registry.mjs
```diff
@@ -84,6 +84,33 @@ const RULE_IDS_TO_SKIP_REGISTRATION = new Set([
...
 ]);
 
+// Fine-grained category → the clear, user-facing bucket the scan output
+// groups & labels by. Rules (and the buckets below) declare a detailed
+// category for intent; the reporter only ever shows these five outcome
+// buckets, so "is this a bug, a slowdown, a vulnerability, an a11y gap,
+// or a maintainability smell?" is obvious at a glance. Collapsing happens
+// here at codegen so every consumer (renderer, JSON, severity overrides,
+// explain) reads the same bucket off `rule.category`.
+const CATEGORY_BUCKET = {
```
aidenybai, line 94: this is unnecsary aliasing that causes inssues, you hsould jusrt fully migrate/refactor this instead of havin gthis weird in-between

packages/oxlint-plugin-react-doctor/src/plugin/rules/architecture/react-compiler-destructure-method.ts
```diff
@@ -69,10 +69,11 @@ const buildHookBindingMap = (componentBody: EsTreeNode | null | undefined): Map<
 // We don't fire when the binding is destructured already.
```
aidenybai: unnecessary comment

packages/react-doctor/src/cli/utils/launch-agent.ts
```diff
@@ -0,0 +1,67 @@
...
+// Each agent's "auto-run / skip-approval" flag. We hand off so the agent can
+// FIX the issues end-to-end; stopping to confirm every edit & command would
+// defeat the point, so we launch in each agent's bypass-approvals mode:
+//   claude  → --dangerously-skip-permissions
+//   codex   → --yolo (bypass approvals + sandbox)
+//   cursor  → --force (auto-approve commands; `--yolo` is its alias)
+// The user already opted in by picking the agent from the handoff menu.
+export const CLI_AGENT_AUTO_FLAGS = {
+  "claude-code": ["--dangerously-skip-permissions"],
+  codex: ["--yolo"],
```
aidenybai, line 28: this is ok

packages/core/src/run-inspect.ts
aidenybai: Keeping `node:path` here — it's imported directly in ~38 core files and there's no `effect/platform` Path/FileSystem service anywhere in the codebase yet. Adopting it in one spot would be inconsistent; a platform-service migration is worth doing as its own dedicated pass.

packages/core/src/runners/oxlint/parse-output.ts
aidenybai: Same as the run-inspect thread — staying on `node:path` for consistency with the rest of core until a dedicated `effect/platform` migration.

packages/core/src/project-info/constants.ts
aidenybai: Added `umd` and `min` in 2a0b6f0f — pattern is now `/.(iife|umd|global|min).js/i`, with test coverage and over-match guards (e.g. `admin.js`, `format.minified.js` stay lintable).

packages/core/src/utils/is-large-minified-file.ts
aidenybai: Right — the sniff already keys on longest-line length (with an avg-line guard so one long line doesn't false-flag). Keeping it split: `isMinifiedSource` is reused standalone by `parse-output.ts` (content sniff, no size gate) and has its own tests; `isLargeMinifiedFile` is just the size-gate wrapper for tree discovery.

packages/core/src/validate-config-types.ts
aidenybai: Renamed to `validatedSurfaceControls` in 2a0b6f0f.

packages/oxlint-plugin-react-doctor/src/plugin/rules/architecture/react-compiler-destructure-method.ts
aidenybai: Removed in 2a0b6f0f.

packages/oxlint-plugin-react-doctor/scripts/generate-rule-registry.mjs
aidenybai: This is the two-level category system: rules declare a fine-grained category (used by `explain` + for intent) that collapses to one of 5 display buckets here. Removing the in-between means rebucketing 300+ rule files and updating the renderer / JSON / explain / severity-override consumers — too large for this PR's scope. Tracking as a dedicated follow-up refactor; leaving this open.

packages/core/src/utils/is-minified-source.ts
```diff
@@ -0,0 +1,36 @@
+import fs from "node:fs";
```
aidenybai, line 1: should this use fs node service with efect or nah

aidenybai: Same call as the `node:path` threads — keeping `node:fs` direct. There's no `effect/platform` FileSystem service anywhere in core yet (every IO util uses `node:fs`/`node:path` directly), and this sniff runs during whole-tree discovery where a synchronous `statSync`/`openSync` is the simplest fit. Worth revisiting as part of a dedicated `effect/platform` migration rather than one-off here.

NisargIO:
lgtm!

packages/core/src/run-inspect.ts
aidenybai: Correction to my earlier reply — `git.ts` *does* use `@effect/platform-node-shared` (NodePath / NodeFileSystem), so the pattern exists. The reason it's there: git needs `ChildProcessSpawner` for streaming spawns, and the FS/Path layers ride along. Here in `run-inspect.ts`, `path` is only pure string join/resolve (no IO) — the `Path.Path` service buys nothing behaviorally and would just add layer wiring to the runInspect program. The dominant pattern for pure path work in core (47 files + the `Files` service itself, which imports `node:path` directly) is raw `node:path`, so I'm keeping it. Happy to migrate if you'd rather standardize on the platform service everywhere.

packages/core/src/runners/oxlint/parse-output.ts
aidenybai: Same correction as the run-inspect thread: the platform pattern does exist (git.ts). But `parse-output.ts` only uses `path.resolve`/`isAbsolute` (pure, no IO), so raw `node:path` matches the `Files`-service / dominant-core pattern. No functional or testability gain from the `Path.Path` service here.

packages/core/src/utils/is-minified-source.ts
aidenybai: Correcting my earlier reply — the platform FS service does exist in core (git.ts uses NodeFileSystem). I'm still keeping raw `node:fs` here specifically because this is a *synchronous* leaf helper doing a bounded `openSync`/`readSync` prefix read during sync file discovery (`listSourceFiles`/`countSourceFiles`). The Effect `FileSystem` service is async and doesn't expose a clean bounded-prefix read, and threading Effect through the sync discovery path would be a large change for no behavioral gain. If we ever make discovery Effect-native, this is the spot to switch.

packages/react-doctor/src/cli/utils/compute-score-projection.ts
```diff
@@ -0,0 +1,40 @@
...
+    TOP_ERRORS_DISPLAY_COUNT,
+    buildRulePriorityMap([currentScore]),
+  );
+  if (topErrorRuleKeys.size === 0) return null;
+
+  const remainingDiagnostics = rescoreSource.filter(
+    (diagnostic) => !topErrorRuleKeys.has(`${diagnostic.plugin}/${diagnostic.rule}`),
+  );
+  const potentialScore = await calculateScore(remainingDiagnostics);
+  if (!potentialScore || potentialScore.score <= currentScore.score) return null;
```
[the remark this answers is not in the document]
aidenybai: False positive — `calculateScore` can't reject. It wraps the entire fetch in try/catch and returns `null` on every failure path: non-OK responses (calculate-score.ts:100-103) and any thrown/abort/timeout error (106-108). So `await calculateScore(...)` always resolves, `computeProjectedScore` returns `null` on failure (line 37 handles falsy), and the surrounding `Effect.promise` never sees a rejected promise — no defect, no crash. The rest of `computeProjectedScore` is synchronous/pure. Adding `orElseSucceed`/try-catch here would be dead code, so leaving as-is.

packages/react-doctor/src/cli/utils/render-multi-project-summary.ts
```diff
@@ -113,10 +151,24 @@ export const printMultiProjectSummary = (input: MultiProjectSummaryInput): Effec
...
+    // project's diagnostics moves the aggregate (its score IS the total).
+    const potentialScore = lowestScoredScan
+      ? yield* Effect.promise(() =>
+          computeProjectedScore(
+            surfaceDiagnostics,
+            lowestScoredScan.result.diagnostics,
+            lowestScoredScan.result.score,
+          ),
+        )
+      : null;
```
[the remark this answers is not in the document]
aidenybai: Same as the compute-score-projection thread: `calculateScore` already swallows all failures and returns `null` (calculate-score.ts:100-108), so this `Effect.promise` can't receive a rejected promise. No recovery needed.

packages/core/src/run-inspect.ts
```diff
@@ -245,6 +293,7 @@ export const runInspect = <HooksR = never>(
       userConfig: resolvedConfig.config,
       readFileLinesSync: fileReader(filesService, scanDirectory),
       respectInlineDisables: input.respectInlineDisables,
+      showWarnings: input.warnings ?? resolvedConfig.config?.warnings ?? DEFAULT_SHOW_WARNINGS,
```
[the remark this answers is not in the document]
aidenybai: Valid — fixed in 4635da3e. Confirmed the per-element pipeline runs over the dead-code stream too, so with warnings hidden `deadCodeCollected` was always empty (never reaching diagnostics or the score) while the worker still spawned. Gated `shouldRunDeadCode` on the resolved `showWarnings` so the default path skips the pass; `--warnings` / `"warnings": true` / `--fail-on warning` re-enable it. (The confusing dead-code *failure* message is gone too, since the pass no longer runs.)

packages/core/src/run-inspect.ts
```diff
@@ -319,7 +370,14 @@ export const runInspect = <HooksR = never>(
...
 
-    const shouldRunDeadCode = input.runDeadCode && !isDiffMode;
+    // Dead-code analysis only ever emits `"warning"`-severity diagnostics
+    // (the `deslop` plugin, all `Maintainability`). When warnings are
+    // hidden that output is filtered out before it reaches any surface or
+    // the score, so the expensive pass (separate worker, large heap, long
+    // timeout) would be pure wasted work — skip it. `--warnings` /
+    // `warnings: true` (and `--fail-on warning`, which forces warnings on)
+    // re-enable it.
+    const shouldRunDeadCode = input.runDeadCode && !isDiffMode && showWarnings;
```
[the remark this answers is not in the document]
aidenybai: Valid — fixed in fc9ed91b. The gate now runs dead-code when warnings are hidden AND a severity override would surface its findings: `deadCodeMaySurfaceWhenWarningsHidden()` checks the `Maintainability` category and any `deslop/*` rule for a `warn`/`error` override (reusing the same controls the pipeline restamps with). Centralized `DEAD_CODE_PLUGIN`/`DEAD_CODE_CATEGORY` so the check can't drift from the emitted diagnostics, and added unit coverage for the override cases.

packages/react-doctor/src/cli/utils/render-diagnostics.ts
```diff
@@ -80,6 +41,10 @@ interface CategoryDiagnosticGroup {
   ruleGroups: [string, Diagnostic[]][];
 }
 
+// Resolves the absolute project root a given diagnostic's relative
+// `filePath` should be read from when building its inline code frame.
+type SourceRootResolver = (diagnostic: Diagnostic) => string;
```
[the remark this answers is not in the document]
aidenybai: Fixed in fc9ed91b — converted to `interface SourceRootResolver { (diagnostic: Diagnostic): string }` per AGENTS.md.

PR #621, opened 2026-06-01
feat(cli): add Sentry crash reporting

NisargIO:
## Summary
- Adds opt-out Sentry crash reporting to the `react-doctor` CLI via `@sentry/node`. Errors reaching the CLI's error funnels (top-level `parseAsync().catch`, the `inspect` and `install` command catches) are captured and flushed before the process exits.
- Each event carries a `run` context snapshot — `version`, `node`/`platform`/`arch`, the invocation `command`/`argv`, `cwd`, CI provider, coding agent, interactivity, and JSON mode — plus a few filterable tags (`origin`, `command`, `ciProvider`, `codingAgent`) to make crashes triage-able.
- Sentry initializes as the **first statement** of the CLI entry (`src/instrument.ts` → `initializeSentry()` in `src/cli/index.ts`) so its global `uncaughtException`/`unhandledRejection` handlers are armed before any command runs. It's scoped to the CLI only — the programmatic `@react-doctor/api` library never initializes Sentry.
- `@sentry/node` is kept external in the bundle (added to `vite.config.ts` `neverBundle`, same rationale as `effect`/`deslop-js`).

## Opt out / safety
- `--no-score` disables crash reporting along with the hosted score API and share URL.
- Sentry is skipped under this repo's own test runs (`VITEST` / `NODE_ENV=test`) so the e2e subprocess suite never phones home.
- `reportErrorToSentry` early-returns when Sentry isn't initialized and swallows transport failures so telemetry can never mask the user's original error.

## Notes
- The opt-in OpenTelemetry/OTLP exporter feature is unchanged.

## Test plan
- [x] `pnpm typecheck` (8/8)
- [x] `pnpm format:check`
- [x] CLI build (`@sentry/node` external, init survives tree-shaking)
- [x] `handle-error`, `install-action`, `json-mode`, `strip-unknown-cli-flags` tests pass
- [ ] Confirm a real crash shows up in the Sentry project with the `run` context attached
- [ ] Confirm `--no-score` suppresses the event


packages/react-doctor/src/cli/utils/build-run-context.ts
```diff
@@ -0,0 +1,113 @@
...
+  ["CLAUDECODE", "claude-code"],
+  ["CLAUDE_CODE", "claude-code"],
+  ["CURSOR_AGENT", "cursor"],
+  ["CODEX_CI", "codex"],
+  ["CODEX_SANDBOX", "codex"],
+  ["CODEX_SANDBOX_NETWORK_DISABLED", "codex"],
+  ["OPENCODE", "opencode"],
+  ["GOOSE_TERMINAL", "goose"],
+  ["AMP_THREAD_ID", "amp"],
+];
```
rayhanadev: this is duplicated with the is-ci-environment.ts file. please consolidate the implementations.

you should have your agent run `truffler`/find-similar-symbols skill as part of the deslop process. 

run /deslop on this pr

rayhanadev (changes requested):
/deslop but wiring lgtm

packages/react-doctor/src/cli/utils/build-run-context.ts
NisargIO: Consolidated in d1b4ef72. Moved the CI-provider and coding-agent label maps plus `detectCiProvider`/`detectCodingAgent` into `is-ci-environment.ts` as the single source of truth; the flat agent env-var list is now derived from the brand map and `isCodingAgentEnvironment` derives from `detectCodingAgent`. `build-run-context.ts` just imports the resolvers now. Behavior is unchanged — the `is-ci-environment` tests still pass.

PR #523, opened 2026-05-27
feat(oxlint-plugin): add 10 performance rules derived from 3perf, coryhouse, million-lint-benchmark

aidenybai:
## Summary

Adds **10 new rules** derived from a deep mining pass over five performance resources:

1. The original 3perf "Unnecessary Renders / Expensive Renders & Effects / Chains of Rerenders / Hydration" catalog (slides + workshop fixtures)
2. [millionjs/million-lint-benchmark](https://github.com/millionjs/million-lint-benchmark) before/after pairs
3. [epicweb-dev/react-performance](https://github.com/epicweb-dev/react-performance) — all 7 exercises
4. [coryhouse/reactjsconsulting#77](https://github.com/coryhouse/reactjsconsulting/issues/77) — "42+ ways to make your React app faster"
5. [3perf/react-workshop-ra](https://github.com/3perf/react-workshop-ra) — workshop fixture app
6. [kurtextrem/awesome-performance-patches](https://github.com/kurtextrem/awesome-performance-patches) (general performance, mostly non-React)

All ten rules target codebases that **do not** use React Compiler — the four `jsx-no-new-*-as-prop` rules + `jsx-no-jsx-as-prop` already use `disabledBy: ["react-compiler"]` for the same reason, and the new memoisation-flavoured rules here join them.

## The 10 rules

| # | Rule | Severity | Category | `disabledBy: react-compiler` | Source |
|---|---|---|---|---|---|
| 1 | `redux-useselector-returns-new-collection` | warn | Performance | ✓ | 3perf #5 |
| 2 | `redux-useselector-inline-derivation` | warn | Performance | ✓ | million-benchmark `ActiveAuthors` |
| 3 | `no-create-context-in-render` | error | Correctness | — | epicweb-dev/02.context |
| 4 | `no-create-store-in-render` | error | Correctness | — | zustand/jotai/redux/mobx/valtio/nanostores |
| 5 | `prefer-stable-empty-fallback` | warn | Performance | ✓ | 3perf #4 |
| 6 | `rerender-lazy-ref-init` | warn | Performance | — | obvious mirror of useState rule |
| 7 | `no-effect-with-fresh-deps` | error | State & Effects | — | exhaustive-deps companion |
| 8 | **`no-random-key`** | error | Correctness | — | coryhouse "Assure keys are assigned, and stable over time" |
| 9 | **`prefer-module-scope-pure-function`** | warn | Architecture | — | coryhouse + 3perf workshop |
[...]

Issue #404, opened 2026-05-22
Dead-code phase fails on all platforms: native bindings missing from published tarball + undeclared transitive deps


SiTaggart:
## Summary

In `react-doctor@0.2.2`, the dead-code phase (powered by bundled `deslop-js@0.0.11`, which uses `oxc-parser@0.132.0` and `oxc-resolver@11.19.1`) fails to load its native bindings on every fresh install, regardless of package manager (npm/pnpm/pnpm dlx). The scan continues with `dead-code` in `skippedChecks` and a `"Cannot find native binding"` reason on `skippedCheckReasons["dead-code"]`, so the score is silently incomplete.

There are **two compounding upstream bugs**, both in react-doctor's package, not in `deslop-js` / `oxc-parser` / `oxc-resolver`.

## Bug 1: `oxc-parser` and `oxc-resolver` are not declared dependencies

`react-doctor`'s `package.json` declares:

```json
"dependencies": {
  "agent-install": "...",
  "oxlint": "...",
  "oxlint-plugin-react-doctor": "...",
  "prompts": "...",
  "typescript": "..."
}
```

No `deslop-js`, no `oxc-parser`, no `oxc-resolver`. The bundled `dist/cli.js` does `require('@oxc-parser/binding-<platform>')` and `require('@oxc-resolver/binding-<platform>')`, but because neither parent package is declared, the platform-specific bindings (which are optionalDependencies of `oxc-parser` / `oxc-resolver`) never get installed.

Repro:
```bash
mkdir /tmp/rd-repro && cd /tmp/rd-repro && npm init -y && npm i react react-dom react-doctor@0.2.2 -D
npx react-doctor . --json --offline | jq '.projects[0].skippedCheckReasons'
```

Output (on darwin-arm64):
```
"dead-code": "Cannot find native binding. ... Cannot find module '@oxc-parser/binding-darwin-arm64' ..."
```

## Bug 2: Rollup-fingerprinted `.node` assets are excluded by `files`

Even if the parent `oxc-*` packages are added as a sibling devDep, the bundle's loader chain doesn't fall through to the standard `@<scope>/binding-<platform>` package lookup. Rollup rewrote the NAPI loader to look for fingerprinted local assets first:

```
./assets/resolver.darwin-arm64-DG19oi97.node   ← Rollup-fingerprinted, not in tarball
./resolver.darwin-arm64.node                   ← not in tarball
@oxc-resolver/binding-darwin-universal         ← package doesn't exist
[...]

PR #630, opened 2026-06-01
fix(oxlint-plugin-react-doctor): declare oxc-parser as a runtime dependency

### LINKED ISSUE #629: react-doctor@latest may fail before scanning with missing oxc-parser
Hi, thanks for maintaining React Doctor.

I may be missing something, but the quick-start style `react-doctor@latest` invocation appears to fail before scanning in my environment.

The README recommends running the latest package directly, for example:

```sh
npx react-doctor@latest
```

In my pnpm-based setup, I reproduced the issue with the equivalent pnpm executor:

```sh
pnx react-doctor@latest --verbose
```

This resolves to `react-doctor@0.2.14` for me.

## What happened

The command fails before the scan starts:

```text
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'oxc-parser' imported from .../node_modules/oxlint-plugin-react-doctor/dist/index.js
Did you mean to import "oxc-parser/src-js/index.js"?
```

## What I expected

I expected the direct `react-doctor@latest` invocation to start scanning the project, or at least fail later with diagnostics from the target project.

## What I found while checking

From the installed package, `oxlint-plugin-react-doctor@0.2.14/dist/index.js` appears to import `oxc-parser`:

```js
import { parseSync } from "oxc-parser";
```

But the published `package.json` for `oxlint-plugin-react-doctor@0.2.14` appears to list `oxc-parser` under `devDependencies`, not `dependencies`.

That makes me suspect the latest published package may rely on a dependency that is available during development/build, but not installed for consumers under strict package-manager resolution.

## Environment

- `react-doctor`: `0.2.14`
- `oxlint-plugin-react-doctor`: `0.2.14`
- Node: `v24.14.1`
- Package manager: pnpm / pnx
- OS: Linux / WSL

## Possible fix

If this diagnosis is correct, moving `oxc-parser` from `devDependencies` to `dependencies` in `oxlint-plugin-react-doctor` may fix this for consumers.

I also noticed #404, but that looked like a previous dead-code/native-binding packaging issue fixed in `react-doctor@0.2.3`. This seems similar in shape, but it appears to involve `oxlint-plugin-react-doctor@0.2.14` directly.

rayhanadev:
## Summary

Fixes #629 — `react-doctor@latest` (0.2.14) can crash **before scanning** with:

```text
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'oxc-parser' imported from
.../node_modules/oxlint-plugin-react-doctor/dist/index.js
```

`oxlint-plugin-react-doctor`'s published `dist/index.js` performs a **top-level static** `import { parseSync } from "oxc-parser"` (cross-file parsing for rules like `no-mutating-reducer-state`). `oxc-parser` is intentionally kept external at build time (`vite.config.ts` → `neverBundle: ["oxc-parser"]`, because its NAPI binding can't be bundled — same class as #404), so the published bundle needs it present at runtime. But it was only declared under `devDependencies`, which consumers never install.

This is a **phantom (undeclared) runtime dependency**. It's normally masked because `react-doctor → deslop-js → oxc-parser` supplies a copy that pnpm's default `hoist=true` exposes to the plugin. The bug surfaces when that accidental satisfaction goes away — strict installs (`hoist=false` / restrictive node-linker configs), other package managers/layouts, or any future `deslop-js` that drops/relocates `oxc-parser`.

## Fix

- Move `oxc-parser` from `devDependencies` → `dependencies` in `packages/oxlint-plugin-react-doctor/package.json` (kept `^0.132.0`, still external/not bundled).
- Update `pnpm-lock.yaml` (clean move, no version drift).
- Add a changeset. The three published packages (`react-doctor`, `eslint-plugin-react-doctor`, `oxlint-plugin-react-doctor`) are a `fixed` group, so this patch-bumps all three; the ESLint plugin also consumes the oxlint plugin and inherits the fix.

## Regression guard (new)

Added `scripts/check-published-deps.ts` + `pnpm check:published-deps` to prevent this whole class from recurring (it slipped in via #523 — a new runtime `import` without a manifest update). The guard:

- Parses every **published** package's built `dist`/`bin` output (TypeScript AST) and collects external `import` / `export … from` / dynamic `import()` / `require()` specifiers.
- Fails if any specifier resolves to a package **not** declared in that package's `dependencies` / `peerDependencies` / `optionalDependencies`.
- Is wired into the `release` script (before `changeset publish`) and into CI on PRs.

I verified it both ways: it passes on the fixed tree, and (temporarily reverting the fix) it fails with a pointed message naming `oxc-parser` and the importing file.

## Audit result (whole publish surface)

Beyond #629, I audited all three published packages' built output (published `0.2.14` **and** a fresh build of this branch). **`oxc-parser` was the only phantom dependency.** `react-doctor` and `eslint-plugin-react-doctor` are clean (e.g. `picomatch` is bundled inline into `react-doctor`, not external).

## Reproduction & verification (against published 0.2.14)

Default `pnpm add react-doctor@0.2.14` → plugin loads fine (phantom satisfied by `deslop-js` + hoisting). With `hoist=false` in `.npmrc`:

- **Before:** `react-doctor --verbose` exits 1 with the exact `Cannot find package 'oxc-parser'` error, before scanning.
- **After** (published tarball with only the dependency move applied, installed via override into the same strict tree): `oxc-parser` resolves from the plugin's own dep tree, the plugin imports OK, and the CLI proceeds to normal project discovery.

## Test plan

- [ ] CI green (lint / typecheck / tests / format / `smoke:json-report` / new `check:published-deps`)
- [ ] On release, `pnpm dlx react-doctor@<next>` under a strict `hoist=false` install starts scanning instead of crashing

---

> [!NOTE]
> **Low Risk**
> Packaging and release-guard changes only; no application logic, auth, or data-handling behavior is modified.
> 
> **Overview**
> Fixes **#629** where `react-doctor` and related plugins could crash before a scan with `ERR_MODULE_NOT_FOUND` for **`oxc-parser`** on strict installs (e.g. pnpm with `hoist=false`). Published **`oxlint-plugin-react-doctor`** keeps **`oxc-parser`** external at build time but only listed it under **`devDependencies`**; it is now a real **`dependencies`** entry (lockfile updated, changeset added).
> 
> Adds **`scripts/check-published-deps.ts`** and **`pnpm check:published-deps`**, which parse each published package’s **`dist`/`bin`** output and fail if any runtime import is not declared in **`dependencies`**, **`peerDependencies`**, or **`optionalDependencies`**. The check runs in **CI** (after **`pnpm build`**) and in the **`release`** script before publish.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1251308ff810b8041ecd04f6ab4131719f29aa71. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/fix-oxc-parser-runtime-dependency.md
+++ b/.changeset/fix-oxc-parser-runtime-dependency.md
@@ -1,0 +1,5 @@
+---
+"oxlint-plugin-react-doctor": patch
+---
+
+Fix `react-doctor@latest` (and the ESLint/oxlint plugins) crashing before the scan starts with `ERR_MODULE_NOT_FOUND: Cannot find package 'oxc-parser'` under strict package managers like pnpm. The published `oxlint-plugin-react-doctor/dist/index.js` performs a runtime `import { parseSync } from "oxc-parser"` (cross-file parsing for rules like `no-mutating-reducer-state`) and the build intentionally keeps `oxc-parser` external, but the package only declared it under `devDependencies`, so consumers never had it installed. `oxc-parser` is now a real `dependency`. See #629.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,6 +91,16 @@ jobs:
         if: ${{ matrix.os == 'ubuntu-latest' && matrix.node-version == '22.18.0' }}
         run: pnpm smoke:json-report
 
+      # Regression guard for #629: a published package's built output must not
+      # import a package that isn't in its `dependencies` (such phantom imports
+      # only resolve when a sibling dependency happens to supply them, and crash
+      # consumers on strict installs). Builds every package, then audits dist.
+      - name: Check published packages declare their runtime dependencies
+        if: ${{ matrix.os == 'ubuntu-latest' && matrix.node-version == '22.18.0' }}
+        run: |
+          pnpm build
+          pnpm check:published-deps
+
       # Allocates a real pseudo-terminal (`pty.openpty()`) so the CLI sees an
       # interactive TTY and renders the multiselect prompt, then asserts the
       # process stays alive waiting for input instead of exiting by itself
--- a/package.json
+++ b/package.json
@@ -23,7 +23,8 @@
     "check": "vp check",
     "changeset": "changeset",
     "version": "changeset version",
-    "release": "pnpm build && node scripts/sentry-sourcemaps.mjs && changeset publish",
+    "release": "pnpm build && pnpm check:published-deps && node scripts/sentry-sourcemaps.mjs && changeset publish",
+    "check:published-deps": "node --experimental-strip-types --no-warnings scripts/check-published-deps.ts",
     "smoke:json-report": "node --experimental-strip-types --no-warnings scripts/smoke-json-report.ts",
     "smoke:tty-prompt": "python3 scripts/smoke-tty-prompt.py",
     "build:schema": "node --experimental-strip-types --no-warnings scripts/generate-config-schema.ts"
--- a/packages/oxlint-plugin-react-doctor/package.json
+++ b/packages/oxlint-plugin-react-doctor/package.json
@@ -54,11 +54,11 @@
   "dependencies": {
     "@typescript-eslint/types": "^8.59.3",
     "eslint-scope": "^9.1.2",
-    "eslint-visitor-keys": "^5.0.1"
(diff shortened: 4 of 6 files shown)


rayhanadev:
bugbot run

PR #739, opened 2026-06-08
chore(ci): compress comments in generated workflow template

NisargIO:
## Summary

- Collapses the multi-line YAML comments in the CLI-generated `.github/workflows/react-doctor.yml` to a single explanatory sentence per block (triggers, permissions, concurrency). New users no longer scroll past five-line paragraphs to read what each block does.
- Behavior is unchanged — same triggers (`pull_request`, `push: main`), same four permissions (`contents: read`, `pull-requests: write`, `issues: write`, `statuses: write`), same `millionco/react-doctor@v2` ref, same concurrency group, same `with:` configuration knobs.
- Adds a `patch` changeset.

### Before

```yaml
permissions:
  # `actions/checkout` needs this to read the repo source.
  contents: read
  # Two uses: (1) reads the PR's changed-file list so the scan only checks
  # what the PR touched (faster, scoped to the diff), and (2) posts/updates
  # the sticky React Doctor summary comment on the PR. Downgrade `write` to
  # `read` to keep the changed-file scan but disable comment posting.
  pull-requests: write
  # The sticky-comment step uses GitHub's `issues.createComment` /
  # `issues.updateComment` endpoints — those are the same APIs that back PR
  # comments (PRs are issues under the hood). Not exercised on `push`
  # events, so safe to drop if you only run on `main`.
  issues: write
  # Lets the action publish a commit status with the score + error/warning
  # counts (links to the run). This is how a `push` to `main` surfaces its
  # result, since the PR comment is skipped off pull requests. Drop it to
  # disable the status (or set `commit-status: false` below).
  statuses: write
```

### After

```yaml
# Needed to check out the repo, scope the scan to the PR's changed files, post/update the sticky summary + inline review comments, and publish the commit status with the score.
permissions:
  contents: read
  pull-requests: write
  issues: write
  statuses: write
```

Same compression applied to `on:` (one line per trigger) and `concurrency:` (one line total). The matching docs example on the website (`apps/website-rd/.../github-actions-setup.mdx`) is updated in [react-review](https://github.com/millionco/react-review) so the rendered docs and the generated workflow stay in sync.

## Test plan

- [x] `pnpm vp test run install-react-doctor action-upgrade` (37/37 pass) — asserts the generated content still contains `name: React Doctor`, all four permissions, `actions/checkout@v5`, `millionco/react-doctor@v2`, and does not contain `github-token` / `diff: main`.
- [x] `pnpm typecheck`
- [ ] Eyeball the generated `react-doctor.yml` in a fresh project after `npx react-doctor` to confirm it reads cleanly.

### DIFF
--- a/packages/react-doctor/src/cli/utils/install-github-workflow.ts
+++ b/packages/react-doctor/src/cli/utils/install-github-workflow.ts
@@ -6,13 +6,11 @@ export interface InstallGitHubWorkflowResult {
   readonly workflowPath: string;
 }
 
-// Self-documenting workflow file. The inline YAML comments walk a new user
-// through the three things they need to change first (non-blocking rollout,
-// scanning `main` on every push for a quality-trend graph, suppressing PR
-// comments) and explain why each permission is granted — without forcing
-// them off to the docs site to learn the basics. The action itself is pinned
-// to the floating major `@v2` (never `@main`, per the supply-chain guidance
-// in AGENTS.md): `@main` would run whatever HEAD points to with
+// Self-documenting workflow file. The inline YAML comments give a new user
+// one-line explanations of each trigger and the permissions block, without
+// forcing them off to the docs site to learn the basics. The action itself
+// is pinned to the floating major `@v2` (never `@main`, per the supply-chain
+// guidance in AGENTS.md): `@main` would run whatever HEAD points to with
 // `pull-requests: write` granted.
 const buildWorkflowContent =
   (): string => `# React Doctor — finds security, performance, correctness, accessibility,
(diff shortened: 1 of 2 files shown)


packages/react-doctor/src/cli/utils/install-github-workflow.ts
```diff
@@ -24,37 +22,21 @@ const buildWorkflowContent =
...
-  # Scans \`main\` on every push so you get a health-score trend on the
-  # default branch — useful for tracking the overall number commit-by-commit
-  # and catching regressions that slipped past PR review. PR-specific steps
-  # (the sticky summary comment) are skipped automatically on \`push\` events.
-  # Comment this block out if you only want PR-time scans.
+  # Scans \`main\` on every push to track the health-score trend and catch regressions that slipped past PR review.
   push:
     branches: [main]
 
+# Needed to check out the repo, scope the scan to the PR's changed files, post/update the sticky summary + inline review comments, and publish the commit status with the score.
```
aidenybai: i feel like this is redundant

PR #826, opened 2026-06-15
feat(rules): add 8 mined correctness, performance, a11y, and security rules

aidenybai:
## Why

A 12-cluster mining pass over the design/engineering brain (React docs, Kent C. Dodds, React/web perf, security, deep a11y, modern-web, RN, …) surfaced ~200 candidate rules. They were deduped, cross-corroborated, and flagged against the existing inventory into `docs/rule-candidates-backlog.md`. This PR ships the highest-signal, lowest-noise batch — each rule is syntax/scope-only, grounded in ≥1 (mostly ≥2) sources, and orthogonal to the current rules.

## What changed

8 new rules (with adversarial tests + false-positive regression coverage), one shared util (`has-jsx-spread-attribute`), and the synthesized backlog doc.

### Bugs
- **`no-call-component-as-function`** — calling a component like `Foo(props)` instead of `<Foo />` runs it outside React (breaks hooks/state/memoization). Fires only on strong local evidence: the name is rendered as JSX in-file **or** defined in-file as a PascalCase function returning JSX.
  ```jsx
  // bad
  <ul>{items.map((item) => Row({ item }))}</ul>
  // good
  <ul>{items.map((item) => <Row key={item.id} item={item} />)}</ul>
  ```
- **`no-create-ref-in-function-component`** — `createRef()` in a function component/hook allocates a fresh ref each render. Class methods/fields and PascalCase non-component factories are left alone.
  ```jsx
  // bad
  function App() { const ref = createRef(); return <div ref={ref} />; }
  // good
  function App() { const ref = useRef(null); return <div ref={ref} />; }
  ```
- **`no-async-effect-callback`** — an `async` effect callback returns a Promise React treats as cleanup, causing unmount races.
  ```jsx
  // bad
  useEffect(async () => { setUser(await load(id)); }, [id]);
  // good
  useEffect(() => { const run = async () => setUser(await load(id)); run(); }, [id]);
  ```

### Performance
- **`no-json-parse-stringify-clone`** — `JSON.parse(JSON.stringify(x))` is a slow, lossy clone → `structuredClone(x)`.
- **`img-missing-dimensions`** — a naked `` (no `width`/`height`, no `className`/`style`, no `srcSet`/`sizes`, no spread) causes layout shift (CLS). Conservative by design.
- **`no-img-lazy-with-high-fetchpriority`** — `loading="lazy"` + `fetchPriority="high"` are contradictory on one image.

### Accessibility
- **`dialog-has-accessible-name`** — a `<dialog>` / `role="dialog"` / `role="alertdialog"` with no `aria-label`/`aria-labelledby`/`title` is announced only as "dialog". Skips custom components and spreads.

### Security
- **`auth-token-in-web-storage`** — persisting tokens (`token`/`jwt`/`secret`/`password`/`apiKey`/…) in `localStorage`/`sessionStorage` exposes them to XSS. Matches `setItem` + property/computed assignment; ignores dynamic keys, reads, and non-storage objects.

### Wiring
- Added `has-jsx-spread-attribute` util (consolidates the spread-bail check several rules need).
- Registered the three in-house ported-bucket rules (`a11y`, `react-builtins`) in `RULES_NOT_PORTED_FROM_EXTERNAL` so `customRulesOnly` keeps them, and regenerated `rule-registry.ts` (387 rules).

## False-positive handling (from a self-review pass)
- `no-create-ref-in-function-component`: requires the enclosing function to be a hook **or** a component that returns JSX, so PascalCase factories (`function MakeStore() { return { ref: createRef() } }`) don't fire.
- `img-missing-dimensions`: skips responsive `srcSet`/`sizes` images and anything sized via `className`/`style`/spread.
- `auth-token-in-web-storage`: curated key pattern (no bare `auth`/`session`/`key`), static keys only.

## Test plan
- [x] `vp test run` for all 8 new suites — 69 passing (adversarial valid/invalid + FP regressions)
- [x] `oxlint-plugin-react-doctor` `typecheck` (incl. `gen`) clean
- [x] `vp lint` clean (only pre-existing fixture warnings)
- [x] `vp fmt --check` clean
- [x] `rule-registry.test.ts` + `rule-metadata` invariants satisfied (titles ≤60 chars, no em/en dashes in title/recommendation, valid category buckets)
- [ ] RDE OSS noise sweep (follow-up before merge per HOW_TO workflow-two)

> Note: a few pre-existing upstream-fixture suites (`no-multi-comp`, `jsx-no-new-*-as-prop`, `no-array-index-key`, `hooks-no-nan-in-deps`) fail in the local stale-build env on `main` too (verified by stashing this work); CI's clean install runs them.

---

> [!NOTE]
> **Low Risk**
> Additive static lint rules and docs only; no runtime or auth behavior changes. `auth-token-in-web-storage` may false-positive on aggressively named non-secret keys, but matching is limited to static keys.
> 
> **Overview**
> Adds **seven new `oxlint-plugin-react-doctor` rules** from a mined backlog, each with rule implementation, adversarial unit tests, and registry wiring. **Bugs:** `no-call-component-as-function` (scope-safe), `no-create-ref-in-function-component`, `no-async-effect-callback`. **Performance:** `no-json-parse-stringify-clone`, `no-img-lazy-with-high-fetchpriority`. **Accessibility:** `dialog-has-accessible-name` (intrinsic dialogs only; bails on spreads). **Security:** `auth-token-in-web-storage` for static sensitive keys on `localStorage`/`sessionStorage` writes.
> 
> Also adds shared **`hasJsxSpreadAttribute`** for spread-bail logic, extends **`RULES_NOT_PORTED_FROM_EXTERNAL`** so in-house rules in ported buckets stay visible under `customRulesOnly`, documents the candidate backlog in **`docs/rule-candidates-backlog.md`**, and records a patch **changeset**. **`img-missing-dimensions` is not shipped** (dropped as not statically sound).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2fdf80a56e37013349fb8afff81f0a406d710a7d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aidenybai:
## RDE / OSS noise sweep (Flow 4) + FP hardening

Built the branch CLI and scanned 4 OSS repos (react-use, radix-ui/primitives, excalidraw, mantine; **~2,793 diagnostics total**), filtered to the 7 new rules and manually classified every hit.

| Rule | Hits | Verdict |
| --- | --- | --- |
| `dialog-has-accessible-name` | 2 (excalidraw ColorPicker, FontPicker) | **True positive** — real `<div role="dialog" aria-modal>` with no accessible name |
| `no-create-ref-in-function-component` | 1 (mantine ScrollArea story) | **True positive** — `createRef()` in a function component |
| `no-call-component-as-function` | 6 (mantine `Menu.test.tsx`) | All test render-helpers (`render(Wrapper({...}))`) → **tagged `test-noise`** |
| `no-async-effect-callback` | 0 | quiet on good code |
| `no-json-parse-stringify-clone` | 0 | quiet on good code |
| `no-img-lazy-with-high-fetchpriority` | 0 | quiet on good code |
| `auth-token-in-web-storage` | 0 | quiet on good code |

**False positives after manual review: 0.**

### Changes from this review
- **Dropped `img-missing-dimensions`** — width/height attrs are only one of several valid ways to reserve space (CSS `aspect-ratio`, container sizing), and the authoritative signal lives in stylesheets a static linter can't see. This is a runtime check (Lighthouse `unsized-images`), so a syntax rule false-positives on any CSS-sized image. (Batch is now **7 rules**.)
- **`no-call-component-as-function`**: resolves the callee through its scope **symbol** (shadow-safe) + `test-noise`.
- **`no-create-ref-in-function-component`**: skips a shadowing local `createRef`; requires the enclosing fn to be a hook or a component that returns JSX.
- Added shadow-safety regression tests for both.

PR #1024, opened 2026-07-02
fix: diff scopes silently skip changed .ts files

rayhanadev:
**TL;DR:** `--scope files`, `--scope changed`, `--scope lines`, and `--staged` drop every changed `.ts`/`.js` file before linting, so a changed React hook that a full scan flags comes back with zero diagnostics. This PR keeps changed non-JSX source files only when their content references React (a react/preact import or a hook call). Non-React `.ts`/`.js` changes stay excluded, so diff scopes stay quiet on server and utility code.

## The bug

`computeExplicitLintIncludePaths` (`packages/core/src/explicit-lint-include-paths.ts`) filters explicit include paths with `JSX_FILE_PATTERN = /\.(tsx|jsx)$/`. `run-inspect.ts` applies that filter to every diff scope. The changed-file list itself keeps `.ts` files (`filterSourceFiles`); they are dropped only at this filter, with no warning.

Before and after, in omgovich/react-colorful at `ddd22b1` with a committed change to `src/hooks/useColorManipulation.ts`:

```console
# before
$ react-doctor --scope files --base <base-ref> --json
0 diagnostics for useColorManipulation.ts

# after (matches the full scan for this file)
$ react-doctor --scope files --base <base-ref> --json
6+ diagnostics: no-derived-state, no-event-handler,
no-pass-live-state-to-parent, no-pass-data-to-parent, ...
```

## Why the filter exists, and why this fix keeps it

The JSX-only filter is the noise guard: including every changed `.ts`/`.js` file would fire generic JS rules on non-React server and utility code in PR reviews. The history shows the current state is drift rather than a consistent policy:

- The original filter (`680e7c40`) applied JSX-only filtering to **full-scan output** too.
- #151 deliberately widened the full scan to `SOURCE_FILE_PATTERN` so JS-performance and adopted rules report on plain `.ts`/`.js`. That is why full scans flag the react-colorful hook today.
- #734 extracted the diff-scope filter into its own helper, preserving the pre-#151 JSX-only semantics the full scan had already abandoned.

So full scans report `.ts` findings and diff scopes hide them, including real React footguns in hooks.

## The fix

Changed non-JSX source files are included only when their content is React-relevant:

```ts
export const shouldIncludeExplicitLintPath = (filePath: string, project?: ProjectInfo): boolean =>
  JSX_FILE_PATTERN.test(filePath) ||
  (project?.framework === "nextjs" && NEXT_ENTRY_FILE_PATTERN.test(filePath)) ||
  (SOURCE_FILE_PATTERN.test(filePath) && hasReactRelevantContent(filePath, project));
```

`hasReactRelevantContent` sniffs the first 256 KB for a react/preact import (any subpath, `import`/`require`/dynamic `import()`) or a hook call (`use[A-Z]...(`), so hook files that only compose other custom hooks still qualify. Unreadable files and non-React files stay excluded, which preserves today's behavior for them. JSX/TSX files and Next middleware/proxy entries are kept without any file read.

Behavior summary:

| Changed file | Before | After |
|---|---|---|
| `src/App.tsx` | linted | linted |
| `src/hooks/useColorManipulation.ts` (imports react) | **skipped** | linted |
| `src/hooks/use-theme.ts` (composes custom hooks) | **skipped** | linted |
| `src/server/db.ts` (no React content) | skipped | skipped (noise guard kept) |
| `middleware.ts` in a Next project | linted | linted |

## Tests

- `explicit-lint-include-paths.test.ts`: rewritten with real file fixtures: React `.ts` hook kept, hook-composing file kept, non-React `.ts`/`.js` dropped, unreadable path dropped, no-project dropped, Next entries kept without reads.
- `diagnose-changed-scope.test.ts` (new): React fixture with a `.ts` hook containing a real footgun; `diagnose` with `includePaths: ["src/use-doubled.ts"]` reports the same diagnostics as the full scan for that file. Fails at 0 diagnostics on the unfixed code.
- `run-inspect.test.ts` and `resolve-lint-include-paths.test.ts` are untouched from `main`: non-React file behavior in those flows is unchanged.

## Validation

- `pnpm test`: green except two pre-existing env-file/gitignore tests (`check-expo-project`, `check-security-scan`) that also fail on a clean checkout of `main` on this machine.
- `pnpm typecheck`: 14/14. `pnpm check`: clean (two pre-existing `no-control-regex` warnings).
- Changeset included (`react-doctor` patch).

Found while validating react-doctor against react-bench (context in millionco/react-bench#65). If you would rather have full parity with #151 (lint every changed source file, generic rules included), that is a one-line change from here; this version keeps the diff-scope noise stance intact.

### DIFF
--- a/.changeset/fix-diff-scope-ts-include-paths.md
+++ b/.changeset/fix-diff-scope-ts-include-paths.md
@@ -1,0 +1,5 @@
+---
+"react-doctor": patch
+---
+
+Diff scopes no longer skip changed `.ts`/`.js` files that contain React code. `--scope files`, `--scope changed`, `--scope lines`, and `--staged` filtered explicit include paths to `.tsx`/`.jsx` only, so a changed `.ts` hook that a full scan flags was reported clean (0 diagnostics). Changed non-JSX source files are now included when their content references React (a react/preact import or a hook call). Non-React `.ts`/`.js` changes stay excluded, so diff scopes remain quiet on server and utility code.
--- a/packages/api/tests/diagnose-changed-scope.test.ts
+++ b/packages/api/tests/diagnose-changed-scope.test.ts
@@ -1,0 +1,69 @@
+import * as fs from "node:fs";
+import os from "node:os";
+import * as path from "node:path";
+import { afterAll, describe, expect, it } from "vite-plus/test";
+import { diagnose } from "../src/index.js";
+
+// Regression fixture for the diff-scope bug where explicit include paths were
+// filtered to JSX/TSX only, silently skipping changed .ts files (hooks, utils)
+// that a full scan flags.
+const projectDirectory = fs.mkdtempSync(path.join(os.tmpdir(), "rdc-changed-scope-"));
+
+fs.writeFileSync(
+  path.join(projectDirectory, "package.json"),
+  JSON.stringify({
+    name: "changed-scope-fixture",
+    private: true,
+    dependencies: { react: "^19.0.0", "react-dom": "^19.0.0" },
+  }),
+);
+fs.writeFileSync(
+  path.join(projectDirectory, "tsconfig.json"),
+  JSON.stringify({ compilerOptions: { jsx: "react-jsx", strict: true } }),
+);
+fs.mkdirSync(path.join(projectDirectory, "src"));
+fs.writeFileSync(
+  path.join(projectDirectory, "src", "use-doubled.ts"),
+  `import { useEffect, useState } from "react";
+
+export const useDoubled = (value: number) => {
+  const [doubled, setDoubled] = useState(0);
+  useEffect(() => {
+    setDoubled(value * 2);
+  }, [value]);
+  return doubled;
+};
+`,
+);
+fs.writeFileSync(
+  path.join(projectDirectory, "src", "app.tsx"),
+  `import { useDoubled } from "./use-doubled.js";
+
+export const App = () => <p>{useDoubled(21)}</p>;
+`,
+);
+
+afterAll(() => {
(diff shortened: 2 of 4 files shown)


rayhanadev:
Reworked per the noise concern. I traced the history: the JSX-only filter dates from 680e7c40 where it also applied to full-scan output, but #151 deliberately widened the full scan to report .ts/.js (JS-perf + adopted rules), and #734 carried the old JSX-only semantics into the extracted helper. So diff scopes and full scans currently disagree by drift, and the diff scope hides real React footguns in changed .ts hooks.

The PR now takes the middle path: changed non-JSX source files are included only when their content references React (react/preact import or a hook call, 256 KB sniff). Non-React .ts/.js changes stay excluded, so the review-noise guard is intact. run-inspect and resolve-lint-include-paths tests are back to untouched-from-main to make that explicit. If you'd rather have full #151 parity instead, it's a one-line change from here.

rayhanadev:
Closing per maintainer direction: the sanctioned way to gate on react-doctor is a full scan dumped to JSON and diffed against a baked pristine baseline (what react-bench verifiers do), not diff scopes. Full scans already report .ts/.js findings (#151), so that pattern never hits this filter. The JSX-only include-path filter stays as intentional noise control for diff scopes.

For the record if this resurfaces: the GitHub Action's PR default is scope: changed (--scope changed --changed-files-from ...), which flows through this filter, so a PR changing only a .ts hook reports no introduced issues in Action reviews. The content-gated fix on this branch (ray/fix-diff-scope-ts-include-paths) addresses that without waking generic JS rules on non-React files, if it's ever wanted.

Issue #858, opened 2026-06-18
Action `pr-files` passes repo-relative paths to a subdirectory scan, so baseline reads miss and pre-existing issues are reported as new


teryn-trimble:
## Environment

- Action: `millionco/react-doctor@v2` (same behavior on `main` as of June 2026)
- CLI: `react-doctor@0.5.x` (reproduced on 0.5.1 and 0.5.6)
- Monorepo layout: `directory: UI` (React app lives in a subdirectory, not repo root)

## Summary

When the GitHub Action scans a subdirectory via `directory: UI`, PR baseline mode (`scope: changed`) incorrectly reports every issue in touched files as newly introduced.

The root cause is that the `pr-files` step writes repo-relative paths from the GitHub API (for example `UI/src/...`) into `--changed-files-from`, but the CLI resolves those paths relative to the scan directory, yielding `UI/UI/src/...`.

## Root cause

1. `pulls.listFiles` returns filenames relative to the repository root (for example `UI/src/modules/foo.tsx`).
2. The action writes those paths unchanged to the changed-files list.
3. The CLI runs against `UI/` and treats each path as relative to that directory, producing `UI/UI/src/modules/foo.tsx`.
4. Baseline materialization uses `git show <base>:UI/UI/src/...`, which does not exist in the committed tree.
5. Base content read fails, so `baseTotalCount: 0`, and every finding in changed files is classified as new.

A common workaround (`ln -s . UI/UI`) makes head scans work on disk but does not fix base reads via `git show`, so the false positives persist.

## Reproduction

1. Use a monorepo with the React app under `UI/`.
2. Configure the action with `directory: UI` and `scope: changed`.
3. Open a PR that touches files under `UI/src/` that already have React Doctor findings on the base branch.
4. Observe the blocking scan reports all issues in those files as new.

## Expected behavior

Baseline mode should compare head vs base for the same logical files and report only issues introduced by the PR. Pre-existing issues in changed files should not fail the gate.

## Actual behavior

All issues in changed files are reported as new (`baseTotalCount: 0`), causing false-positive PR failures.

## Evidence

On a real PR touching one file with 35 existing issues:

**Repo-relative paths (`UI/src/...`)**
- `baseTotalCount`: 0
- New issues: 33
- Fixed issues: 0

**Project-relative paths (`src/...`)**
- `baseTotalCount`: 35
- New issues: 0
- Fixed issues: 2

## Suggested fix

[...]

PR #1065, opened 2026-07-04
fix: --staged silently scans nothing in monorepo subdirectories

### LINKED ISSUE #1064: `--staged` silently scans nothing when the project is a subdirectory of the git repo (`git show :<path>` resolves against repo root)
## Environment

- react-doctor 0.7.1 (`npx react-doctor@0.7.1`)
- Monorepo: git root at repo top level, React project in a subdirectory (e.g. `apps/webui`)
- macOS (Darwin 25.1.0), git 2.x — but the bug is platform-independent

## Summary

`react-doctor --staged` silently scans nothing when the scanned project is a **subdirectory of the git repo** (standard monorepo layout). The CLI prints `Scanning N staged files...` followed by `No issues found!`, and the JSON report shows `scannedFileCount: 0` — every staged file is silently skipped.

## Root cause (observed in shipped `dist/cli.js`)

Two git invocations disagree about what paths are relative to:

1. Staged file paths are collected with `git diff --cached --relative ...` run from the project directory → paths are **project-relative**, e.g. `src/Bad.tsx`.
2. Staged content is then read with `git show :src/Bad.tsx`, also run from the project directory. But git resolves `:<path>` specs relative to the **repo root**, not the cwd, so this fails with:

   ```
   fatal: path 'apps/webui/src/Bad.tsx' exists on disk, but not 'src/Bad.tsx'
   (or: path 'apps/webui/src/Bad.tsx' is in the index, but not 'src/Bad.tsx')
   ```

3. The error is swallowed and the file is skipped with no warning (same silent-degradation path as #937), so the scan "succeeds" with zero files.

When the project directory *is* the repo root, both forms coincide and `--staged` works, which is presumably why this wasn't caught.

## Reproduction

```bash
# monorepo: git root with the React project in a subdirectory
mkdir -p repro/apps/webui/src && cd repro && git init

cat > apps/webui/package.json <<'EOF'
{ "name": "webui", "dependencies": { "react": "^19.0.0" } }
EOF

cat > apps/webui/src/Bad.tsx <<'EOF'
export function Bad({ x }: { x: string }) {
  return <div dangerouslySetInnerHTML={{ __html: x }}>child</div>;
}
EOF

git add apps/webui/package.json apps/webui/src/Bad.tsx

cd apps/webui
npx react-doctor@0.7.1 --staged --no-score -y
# → "Scanning 1 staged files..." then "No issues found!"
#   (JSON report: scannedFileCount: 0)

npx react-doctor@0.7.1 --scope files src/Bad.tsx --no-score -y
# → correctly reports the dangerouslySetInnerHTML-with-children error
```

## Expected behavior

`--staged` scans the staged files and reports the same findings `--scope files` does. If a staged file can't be read from the index, a warning should be printed rather than silently skipping it (per #937).

## Actual behavior

All staged files fail the `git show` read and are silently skipped; the run reports success with `scannedFileCount: 0`. This is particularly bad in a pre-commit hook, where it means the gate passes without checking anything.

## Suggested fix

Use git's cwd-relative pathspec form when reading staged content:

```
git show :./src/Bad.tsx
```

(the `:./ ` prefix makes the path cwd-relative), or alternatively prefix the path with the project's repo-root-relative prefix (`git rev-parse --show-prefix`) before passing it to `git show`.

Related: #937 (silent error swallowing in staged discovery — makes this bug invisible), #858 (same class of repo-root vs. project-dir path confusion in the Action's `pr-files` baseline reads).

skoshx:
## Root Cause

When the scanned project is a subdirectory of the git repo (standard monorepo layout), `--staged` silently scanned nothing because:

1. `git diff --cached --relative` returns **project-relative** paths (e.g., `src/Bad.tsx`)
2. But `git show :<path>` resolves paths relative to the **repo root**, not the cwd
3. So the file read failed silently and all files were skipped with `scannedFileCount: 0`

This is the same class of bug as #858 (PR baseline reads) — both stemmed from git's `:<path>` pathspec being repo-root-relative when the operation runs from a subdirectory.

## Fix

Changed `git show :<path>` to `git show :./<path>` in `showStagedContent`. The `./` prefix makes the pathspec cwd-relative, matching how `showRefContent` already works for baseline reads (which has a comment explaining exactly this issue).

## Scope

- **Only affects `--staged` mode** — the fix is in `Git.showStagedContent`, which is only called during staged file materialization
- **Zero impact on normal scans** — regular file discovery and content reading use different code paths
- **All tests pass**, including new regression test that validates `--staged` works from a monorepo subdirectory

## Parity

Parity check is not applicable for this fix because:
- The corpus scan doesn't use `--staged` mode (it scans all files normally)
- This fix only affects the staged file content reading path
- No rule behavior changed

## Testing

- Added integration test that creates a real monorepo structure and validates staged content is read correctly
- Manually tested with the reproduction case from the issue (works correctly now)
- All existing tests pass

Closes #1064

<div><a href="https://cursor.com/agents/bc-98f6b234-d7d1-466c-b6e2-988bee890230"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-98f6b234-d7d1-466c-b6e2-988bee890230"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.changeset/fix-staged-monorepo.md
+++ b/.changeset/fix-staged-monorepo.md
@@ -1,0 +1,10 @@
+---
+"react-doctor": patch
+"@react-doctor/core": patch
+---
+
+fix: --staged now works in monorepo subdirectories
+
+Fixed a bug where `--staged` silently scanned nothing when the project is in a subdirectory of the git repo (standard monorepo layout). The CLI would report "No issues found!" with `scannedFileCount: 0` because staged file paths were resolved incorrectly.
+
+The fix changes `git show :<path>` to `git show :./<path>` to use cwd-relative pathspecs, matching how baseline reads already work.
--- a/packages/core/src/services/git.ts
+++ b/packages/core/src/services/git.ts
@@ -769,7 +769,7 @@ export class Git extends Context.Service<
         showStagedContent: (directory, relativePath, options) =>
           runCommand({
             command: "git",
-            args: ["show", `:${relativePath}`],
+            args: ["show", `:./${relativePath}`],
             directory,
             maxStdoutBytes: options?.maxBufferBytes,
           }).pipe(Effect.map((result) => (result.status === 0 ? result.stdout : null))),
--- a/packages/core/tests/services/staged-files.test.ts
+++ b/packages/core/tests/services/staged-files.test.ts
@@ -144,6 +144,63 @@ describe("StagedFiles.layerNode — Zip-Slip defense (path traversal)", () => {
   });
 });
 
+describe("StagedFiles.layerNode regression — monorepo subdirectory (#1064)", () => {
+  let monorepoRoot: string;
+  let tempDirectory3: string;
+
+  beforeEach(() => {
+    monorepoRoot = fs.mkdtempSync(path.join(os.tmpdir(), "rd-staged-monorepo-"));
+    tempDirectory3 = fs.mkdtempSync(path.join(os.tmpdir(), "rd-staged-materialize-"));
+
+    const projectDir = path.join(monorepoRoot, "apps", "webui");
+    fs.mkdirSync(projectDir, { recursive: true });
+    fs.mkdirSync(path.join(projectDir, "src"), { recursive: true });
+
+    fs.writeFileSync(
+      path.join(projectDir, "package.json"),
+      JSON.stringify({ name: "webui", dependencies: { react: "^19.0.0" } }),
+    );
+    fs.writeFileSync(
+      path.join(projectDir, "src", "App.tsx"),
+      'export function App() { return <div>Hello</div>; }\n',
+    );
+
+    const { spawnSync } = require("node:child_process");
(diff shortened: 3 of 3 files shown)


rayhanadev:
Closing in favor of the PR shipping from `ray/issue-1064-fix` — same root cause and identical `git show :./<path>` fix, plus a regression test and changeset. Thanks for the independent confirmation on the fix! (づ｡◕‿‿◕｡)づ

PR #1189, opened 2026-07-12
fix(rule): preserve stateful RegExp loop semantics (MVP)

aidenybai:
## Problem

`js-hoist-regexp` recommended hoisting RegExp instances with `g` or `y` flags even though their mutable `lastIndex` changes repeated loop results.

## Fix

Compute effective static flags for RegExp call/new forms, including literal inheritance and explicit overrides, and keep general stateful constructions inside the loop. The detector preserves the safe native `String.prototype.replaceAll` case only when:

- `g` is present
- the construction is the direct search argument
- the receiver is proven primitive string
- the global RegExp constructor and relevant String/RegExp prototype methods are not locally replaced

String proof covers literals, explicit string bindings, stable aliases, and the pinned HyperDX `globSync` string-return contract when `withFileTypes` is absent or statically false. Unknown/userland receivers and dynamic options remain quiet.

## Regression coverage

- `g`, `y`, `gy`, templates, duplicates/invalid flags, literal inheritance and explicit overrides
- RegExp call/new, wrappers, TypeScript expressions, shadowed/reassigned globals
- global-object, alias, Object/Reflect, destructuring, delete/update, and prototype mutation paths
- native replaceAll positives and userland/unknown/optional/computed/nested/second-argument negatives
- permanent fuzz corpus and generator coverage

## Validation

Exact head: `92ecb392a1d4e05b8067c9368cacc4ac886aa38e`

- exact runtime: fresh stateful regexes produce repeated true values while unsafe hoisting changes later results
- exact native replaceAll runtime: `g`/`gy` outputs remain identical and lastIndex resets to zero
- monkey-patch runtime witness: fresh `[true,true,true]` vs hoisted `[true,false,true]`
- plugin suite: 549 files, 12,332 passed, 148 skipped
- whole-repo typecheck: 15/15 tasks
- lint, format, diff-check, truffler/deslop: passed
- fuzz corpus: 38 passed
- strict 500-iteration fuzz: passed with target fire coverage
- force-fresh RDE: 100/100 baseline and candidate, zero failures/partials/skips, 15,047→15,047 diagnostics, zero all-rule identity deltas
- exact React Bench: 120/120, zero failures/incomplete scans, equal 109,048 extents, zero added/removed identities
- pinned HyperDX true positive retained: 1→1 at packages/cli/src/sourcemaps.ts:210:9
- reconstructed HyperDX oracle: complete 625/625 extents, target 1→1, zero deltas
- full CI matrix: green
- review threads: zero

## Precision boundary

Stateful RegExp constructions remain quiet unless a native operation is proven to reset state before every use. Unknown receivers, dynamic flags/options, or locally replaced built-ins are treated conservatively.

---

> [!NOTE]
> **Medium Risk**
> Large change to lint rule semantics and AST/scope analysis (~400+ lines); risk is false negatives on hoisting opportunities and false positives if environment-hazard heuristics are wrong, not runtime security.
> 
> **Overview**
> **`js-hoist-regexp`** no longer suggests hoisting loop-built `RegExp` calls when **`g` or `y` flags** make `lastIndex` semantics matter, while still warning on truly stateless static patterns.
> 
> The rule now resolves **effective static flags** (including flags inherited from regex literals), requires the **global `RegExp` constructor**, and skips reporting when the file may have **replaced `globalThis.RegExp`** or **mutated `String`/`RegExp` prototype** hooks that affect `replaceAll` behavior. A narrow carve-out still **flags** global regexes passed as the **direct** `replaceAll` search on **proven native string** receivers (including **`glob`** `globSync` paths when options are statically string-returning).
> 
> **Tests and fuzz**: large regression suite, liveness fixture switched to non-stateful `i` flags, permanent fuzz corpus cases, handler snippets, and **`js-hoist-regexp`** added to verdict-robustness auditing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5bec10cb6842528c153379783b72206bb2b0d550. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rayhanadev (approved):
Adversarial review found 1 proven HIGH + 1 proven LOW finding; both fixed in d341db570.
- HIGH: the whole-file environment kill switch suppressed unrelated stateless RegExp reports when `String.prototype.replaceAll`/`RegExp.prototype` were mutated (e.g. a replaceAll polyfill silenced every js-hoist-regexp report). Now those mutations veto only the stateful replaceAll carve-out; global `RegExp` replacement still vetoes everything. Pinned with regression fixtures (stateless stays flagged; mixed fixture emits exactly one diagnostic).
- LOW: the duplicate-flag branch of `hasValidRegExpFlags` survived mutation deletion with all tests green; added a non-stateful `"ii"` expectPass isolating it.
Deslop in 4cf9af9f4: -123 lines added (799 -> 676), behavior preserved. Full oxlint-plugin suite green (12334 passed), typecheck + lint clean.
Follow-ups (not blocking): coverage for the String(...)-call and template-literal receiver proofs; pin the optional-chaining guard (`text?.replaceAll` stays quiet); `value: null` regex-literal flags fallback (pre-existing on main); default `import glob from "glob"` not recognized by the globSync proof (accepted MVP scope).

PR #1191, opened 2026-07-12
fix(rules): constrain min max sort advice

aidenybai:
## Why

`js-min-max-loop` used to recommend `Math.min(...array)` or `Math.max(...array)` for caller-owned arrays even though `sort()` mutates the receiver. It also recommended scalar rewrites for values such as `NaN`, mixed signed zero, async comparators, and locally monkeypatched `sort`/`Math` builtins, where the replacement can change behavior.

Before:

```ts
const smallest = values.sort((left, right) => left - right)[0];
```

Supported after this change:

```ts
const smallest = [3, 1, 2].sort((left, right) => left - right)[0];
```

The supported form is a disposable, non-empty, dense array literal of statically finite numeric primitives using a canonical synchronous comparator and index `0`. The rule stays quiet when scalar equivalence or builtin integrity is not proven.

## What changed

- Requires a fresh finite numeric array literal and rejects aliases, spreads, holes, coercive values, `NaN`, infinities, mixed signed zero, unsafe argument counts, and nonzero endpoints.
- Rejects async/generator and noncanonical comparators.
- Proves the visible `Math` binding is native and suppresses advice when the program mutates/replaces `Math.min`/`Math.max` or `Array.prototype.sort`, including static aliases and `Object`/`Reflect` mutation APIs.
- Caches the program mutation scan and keeps stress behavior near-linear.
- Adds paired adversarial rule regressions, a permanent FP corpus seed, generator liveness, and verdict-robustness coverage.

## Semantic validation

- 64,761 randomized finite-array comparisons matched exactly under `Object.is` for ascending→`Math.min` and descending→`Math.max` after excluding mixed signed zero.
- Explicit runtime witnesses confirmed that monkeypatching either `Array.prototype.sort` or `Math.min` changes the original/replacement result.
- Exact built CLI repro: candidate reports only the safe literal; baseline additionally reports the async comparator, caller-owned receiver, and locally mutated `Math` case.
- Stress test for 100/500/1,000/2,000 findings: 55/72/95/127 ms.

## Eval results

| Check | Result |
| --- | --- |
| RDE manifest | `/tmp/rde-pr1189-combined-100-success-manifest.json` |
| RDE repos/rootDirs | 100 distinct / 100 scans on each side |
| RDE failures | 0 baseline / 0 candidate |
| RDE target diagnostics | 0 → 0; 0 added, 0 removed |
| React Bench repos | 120 distinct pinned revisions; 120/120 completed |
| React Bench extents | 96,853 files baseline / 96,853 candidate |
| React Bench target diagnostics | 0 → 0; 0 added, 0 removed |
| React Bench parser audit | 5 identical parser diagnostics per side; every affected file manually inspected, with no canonical target shape |
| React Bench oracle solutions | all 747 tracked solution source files; 0 → 0 |

Artifacts:

- `/tmp/minmax-rde-distinct-summary.json`
- `/tmp/minmax-react-bench-direct-results.json`
- `/tmp/minmax-react-bench-summary.json`
- `/tmp/minmax-react-bench-oracle-results.json`

## Test plan

- `nr --filter oxlint-plugin-react-doctor test` — 549 files; 12,375 passed; 148 skipped.
- `nr --filter @react-doctor/fuzz test` — 38 passed; 400 skipped.
- `FUZZ_RULE=js-min-max-loop FUZZ_STRICT=1 FUZZ_ITERATIONS=500 nr fuzz --reporter=verbose --disableConsoleIntercept` — passed; fire coverage 1/1.
- `nr typecheck` — 15 tasks passed.
- `nr lint` — passed with existing warnings only.
- `nr format:check`.
- `nr smoke:json-report`.
- `git diff --check`.
- `nr test` — all rule/core/API/fuzz suites passed; the saturated parallel run tripped two performance-harness timing checks and one stdin lifecycle check. The performance harness then passed 22/22 in isolation, and the stdin lifecycle file passed 2/2 on retry.
- Post-change truffler searches found no reusable or duplicate helper.

## Draft status

Implementation and local/external validation are complete. This remains a draft while the pushed head runs through the full CI and review-thread audit.

---

> [!NOTE]
> **Medium Risk**
> Substantial rule logic changes reduce false positives but may miss previously flagged patterns; behavior is lint-only with no runtime product impact.
> 
> **Overview**
> **`js-min-max-loop`** no longer suggests `Math.min`/`Math.max` for caller-owned or aliased arrays. Advice is limited to **disposable, dense array literals** of statically finite numbers, canonical sync `(a, b) => a - b` / `(b - a)` comparators, and **`[0]`** only (no `[length - 1]` hints).
> 
> The rule **stays silent** when scalar equivalence is unclear: spreads, holes, `NaN`/infinities, mixed `+0`/`-0`, non-numeric literals, async/generator comparators, oversized literals (above **1024** elements, new `MATH_EXTREMUM_SPREAD_MAX_ELEMENT_COUNT`), or shadowed/`Math`/`Array.prototype.sort` mutations detected via a **per-program cached AST walk**.
> 
> Coverage adds adversarial regressions, a fuzz FP corpus seed for caller-owned sort mutation, liveness/verdict-robustness hooks, and a patch changeset.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1986c7fda276f273db23489ad5fadcfa5ab998a1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rayhanadev (approved):
Adversarial review: 12 confirmed findings (4 proven with failing fixtures, 8 coverage/nit); the in-PR ones are all fixed in c848bdf5 + de668458.
Fixed: inert fuzz corpus seed renamed .ts -> .tsx (the loader only matches .tsx/.jsx, so it was never exercised); 8 missing pinning tests added (deleted [x - 1] branch, Math.min++ update branch, Reflect.getPrototypeOf arm, mutation-scan cache-hit path, self global arm, variable-key defineProperty fallback, let/var/class/import Math transparency); MATH_EXTREMUM_SPREAD_MAX_ELEMENT_COUNT rationale comment.
Deslop: -3 net lines (redundant type-guard re-check, dead finiteness re-check, duplicated key lookup).
Tests: oxlint-plugin suite 549 files / 12385 passed, tsc --noEmit clean, fuzz suite green with the seed now loading.
Follow-ups (not blocking, all pre-existing on main): let/var alias mutation of Math bypasses the mutation guard; destructuring/for-of writes to Math.min invisible to the mutation walk; const Math masked by a later interface Math (last-declaration-wins symbol map); two older corpus seeds (no-document-write--shadowed-global.ts, no-eval--globalthis-static-member.js) are also mis-named and never loaded.

PR #1195, opened 2026-07-12
fix(rule): prove propTypes component ownership

aidenybai:
## Why

`no-prop-types` treated any uppercase assignment receiver as a React component. That reports unrelated validation objects such as:

```ts
const Schema = { propTypes: {} }
Schema.propTypes = validators
```

The rule now requires binding-level component evidence while preserving genuine React 19 findings.

## What changed

- Resolves local function, class, alias, and namespace-merged component identity through scope bindings.
- Tracks reassignment timing at each value-capture point, including multi-hop aliases and class base aliases.
- Preserves binding-proven `memo`, `forwardRef`, `useMemo`, React hook-owning, and `styled-components` components.
- Requires render output to reach a function return, and recognizes `children` only from an untouched `props.children` or direct first-parameter object binding.
- Traverses callback JSX only through binding-proven render-preserving operations (`useMemo`, array `map`, and proven lodash `sortBy` results); `some`, `find`, `forEach`, Promise `then`, shadowed methods, and replaced methods stay quiet.
- Invalidates React class, HOC, and styled factory provenance when the relevant property was replaced.
- Tracks those replacements across sibling aliases and constant-computed property keys that share the same runtime object.
- Requires nested binding mutations to be synchronously reachable before the `propTypes` reference; uncalled, timer-deferred, and post-reference mutations do not erase genuine component evidence.
- Resolves returned `let`/`var` bindings with CFG reaching definitions, so nested, dead, branch-exited, post-return, and overwritten JSX writes cannot manufacture component evidence.
- Preserves direct, renamed, and defaulted first-parameter `children` destructuring while rejecting bare positional names and nested object/array bindings.
- Keeps shadowed/userland React APIs, object-returning hook helpers, discarded callback JSX, ambiguous control flow, imported opaque values, uppercase objects, and object spreads quiet.
- Adds paired regressions and permanent fuzz seeds for every confirmed false-positive shape.

## Validation

Frozen candidate: `5ca11ef219fbaa12239a0ede2da021b5c3c70043`  
Baseline: `c5a06bd8c0d5ade974549326b9131fa9b6ae8e6e`

- Exact seven-file repro: baseline reports the uppercase-object, bare positional `children`, and nested-binding FPs plus two real findings; candidate removes all three FPs and retains both real findings at equal extent.
- Focused provenance and shared render-output regressions: 65/65 pass, including unreachable nested writes, branch-exited writes, post-return writes, unconditional overwrites, loop-initializer overwrites, and true conditional JSX assignments.
- Full plugin suite: 550 files; 12,359 pass, 148 skip.
- Strict fuzz: `FUZZ_RULE=no-prop-types FUZZ_STRICT=1 FUZZ_ITERATIONS=500 nr fuzz`; pass with target fire coverage (39 firing programs, 381 executed programs, zero findings in the direct stats audit).
- Fuzz package: 38 pass, 400 opt-in cases skipped.
- Target-only local RDE: 100/100 repositories, 10,522 files on each side, 197 target findings on each side, zero added/removed diagnostics.
- Target-only React Bench: 120/120 repositories, 97,626 files on each side, 532 target findings on each side, zero added/removed diagnostics.
- Reconstructed React Bench oracles: 5/5, 4,876 files on each side, 571 target findings on each side, zero added/removed diagnostics.
- Full repository tests: 14/14 tasks pass; React Doctor package 205 files passed and 2 skipped (2,208 tests passed, 27 skipped).
- Review-fix validation: exact Cursor reproduction restored; 14/14 adversarial provenance audit cases pass; fresh RDE 100/100, React Bench 120/120, and five-oracle comparisons all remain zero-delta; repository typecheck passed 15/15 tasks.
- `nr typecheck`, `nr lint`, `nr format:check`, `nr smoke:json-report`, and `git diff --check` pass.

The RDE, Bench, and oracle scans force React 19 capability and enable only `react-doctor/no-prop-types`; dependency and generated-output directories are excluded recursively. Every intermediate removed diagnostic was manually inspected. The draft first exposed 22 lost `memo`/`forwardRef`/`useMemo` findings, then three hook/null/children oracle findings, then 18 styled/children Bench findings; all were true-positive regressions, fixed with paired coverage, and restored in the final zero-delta runs.

---

> [!NOTE]
> **Medium Risk**
> Touches core semantic utilities used by many rules beyond no-prop-types; behavior changes are heavily tested but misclassification could silence or add diagnostics in edge cases.
> 
> **Overview**
> **`no-prop-types`** no longer treats every PascalCase `propTypes` receiver as a React component. Assignments and static class fields are reported only after **binding-level provenance** (`isProvenReactComponentSymbol` / `isProvenReactClassComponent`): real function/class components, stable aliases, `memo`/`forwardRef`/`useMemo`, hook-owning renderers, and `styled-components`, while skipping validation objects, reassigned bindings, shadowed React/styled factories, and non-components.
> 
> Shared **render-output and CFG analysis** is tightened and threaded through several rules via optional `context.cfg`. `functionContainsReactRenderOutput` now keys off return-path evidence (including CFG-backed `let`/`var` reaching definitions) and limits callback JSX to render-preserving calls (proven `array.map`, React `useMemo`, lodash `sortBy`). Callbacks in `some`/`find`/discarded results and replaced `map`/`memo`/`styled` methods no longer count. **`functionReturnsMatchingExpression`** and related helpers gain write-before and alias-mutation guards.
> 
> Call sites for **`functionContainsReactRenderOutput`** (e.g. giant component, boolean props, call-component-as-function, only-export-components) pass control-flow analysis so behavior stays aligned with the stricter model.
> 
> Coverage adds a large **`no-prop-types` regression suite**, fuzz corpus seeds, snippet pools, and `no-prop-types` in verdict-robustness; the liveness fixture now uses a component that actually renders JSX.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fada8838c55fe8f8409e9376374ae452c764f4ed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rayhanadev (approved):
Adversarial review: 2 proven findings (one shared root cause), both fixed in the pushed commits — functionReturnsMatchingExpression now follows let/var bindings through their initializer and plain-assignment values, restoring render evidence for JSX built into a mutable local and returned (regressed no-prop-types plus the 6 other consumers of functionContainsReactRenderOutput).
Pinning tests added at helper and rule level in both directions; never-returned let-bound JSX stays suppressed.
Deslop: net -31 lines across 7 files, behavior-preserving; hunks superseded by the newer provenance commits were dropped.
Validation: 12,356 oxlint-plugin tests green, eslint mirror green, typecheck clean, after rebasing onto 6b6153149.

Follow-ups (not blocking):
- Pre-existing FP on main: uppercase utility like (node) => node.children proven as a component.
- Member-expression receivers (ns.Panel.propTypes =) are out of scope by design — undocumented coverage boundary.
- The PR's RDE/Bench zero-delta was measured with only no-prop-types enabled; the other consumers of the shared helper were not separately measured.

PR #1197, opened 2026-07-13
fix(rule): ignore module-initialized inline styles

aidenybai:
## Summary

- stop `no-inline-exhaustive-style` from reporting direct JSX style objects allocated once at module initialization
- cover direct module constants, top-level IIFEs, synchronous callbacks nested inside one-shot initializers, and static/computed class fields
- preserve component, hook, class-instance-field, memoized, deferred, and otherwise unknown function findings
- add a permanent valid fuzz seed, generator shapes for quiet and firing paths, and a patch changeset

## Precision boundary

The rule suppresses a style object only when every enclosing function is proven to execute synchronously as part of a top-level initializer. Module class static fields and computed keys execute once; instance-field values remain diagnostics because each construction allocates them again. Unknown callback/component contexts remain conservative.

Two broader intermediate approaches were rejected by evidence: render-name proof removed five real RDE findings, and treating every no-function expression as module-stable hid a class instance-field TP. Both blockers now have paired regressions. All final evidence below was regenerated from `22b264dbbdbb17126febfeb43e9efc1cc6a99c8c`.

## Exact reproduction

- pinned V28 fixture: 3/3 files scanned by both binaries
- baseline: module FP 1, render TP 1, stable-object negative 0
- candidate: module FP 0, render TP 1, stable-object negative 0
- runtime oracle: repeated module accesses share both element and style identity; repeated component calls create fresh element and style identities
- class matrix: instance-field value 1; module static-field value 0; module computed field key 0

## Validation

- focused regressions: 13/13
- oxlint plugin: 549 files, 12,312 passed, 148 skipped
- fuzz corpus: 37 passed
- strict fuzz: 500 iterations, fire coverage 1/1, no findings (extended only the harness timeout under host contention)
- RDE fresh no-cache: 100/100 repos, 0 failures, target 37→37, 0 added, 0 removed
- React Bench exact pinned set: 120/120 repos, 0 failures, 97,695→97,695 files, target 224→224, 0 added, 0 removed
- oracle census: 2 target tasks; each base remains 1→1 and each oracle solution remains 0→0
- root typecheck: 15/15
- root lint: pass (pre-existing warnings only)
- format check and `git diff --check`: pass
- final-head CI: pass; exact head passed all Linux/macOS/Windows and Node 20/22/24/25/26 checks

## Review status

Draft until final-head CI and direct review-thread audits complete.

---

> [!NOTE]
> **Low Risk**
> Lint-rule precision change with broad regression and fuzz coverage; no runtime app behavior, only fewer false positives for module-scoped JSX.
> 
> **Overview**
> **`no-inline-exhaustive-style`** no longer warns on large inline `style` objects that are allocated once at module load instead of on every render.
> 
> The rule bails out when the style expression sits in **one-shot module initialization**: direct module constants, top-level IIFEs, synchronous callbacks inside those initializers (e.g. `[].map` / `Array.from`), and **static** class fields—while still flagging **instance** fields, render bodies, hooks, deferred callbacks (`setTimeout`), and paths through ordinary named functions.
> 
> Implementation adds **`isOneShotModuleInitialization`** (enclosing-function walk via **`executesDuringRender`**, plus **`isInsideInstanceField`** for nested class-expression edge cases) and an early return on the style expression before counting static properties.
> 
> Coverage: a large regression matrix, fuzz corpus seed **`no-inline-exhaustive-style--module-jsx.tsx`**, and snippet pools for stable vs render-time exhaustive styles; patch changeset for **`oxlint-plugin-react-doctor`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e537cd75bfea7341d0a29ad329f4884940f01c3f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rayhanadev (approved):
Adversarial review: 3 proven + 3 confirmed findings, all addressed in the pushed commits (c137dbfee, e537cd75b).
Fixed: isInsideInstanceField now matches AccessorProperty (per-instance `accessor` fields report again) and walks past static class fields (nested class-expression static field FN) — both new tests fail against the unfixed rule.
Coverage added: preserved module-scope contexts (hook / setTimeout / memo / named fn), useMemo render-phase-factory walk, and Array.from arg-1 suppression (direct + const-aliased).
Deslop: −15 lines (test 1 reuses the shared style fixture). Plugin suite 12,320 tests green; typecheck + format clean.
Follow-ups (not blocking): pre-existing FPs — `data["map"](fn)` computed member and named functions called once at module load still report; receiver-blind `.map` suppression on FRP instance methods is an accepted trade-off.

PR #1198, opened 2026-07-13
fix(no-cascading-set-state): retire unsound redraw warning

aidenybai:
## Why

Retires `no-cascading-set-state` because its detector equates synchronous setter calls with React commits. React 16.14, 17.0.2, 18.3.1 legacy/`createRoot`, and 19.2.5 all batch three synchronous setters inside one effect into one follow-up render. A reducer dispatch produces the same commit count, so the warning and recommendation are not actionable.

The reported screenshot is pinned to `anl331/goey-toast@0dafb8ff` (`GooeyToast.tsx:868`), which uses React 19.2.4.

## What changed

- Keeps the rule ID registered as a default-off retired compatibility stub.
- Removes the unsound setter-count detector and stale positive controls.
- Adds registry/liveness/integration retirement coverage and a permanent fuzz regression seed.
- Replaces the README example that explicitly enabled the retired rule.
- Leaves semantic sibling rules (`no-chain-state-updates`, derived-state rules, self-updating effects, and React Compiler set-state-in-effect) responsible for their narrower proven contracts.

## Exact validation

| Check | Result |
| --- | --- |
| Focused registry/rule/liveness | 471/471 |
| Full oxlint plugin | 12,262 passed, 148 skipped |
| Fuzz corpus | 36 passed |
| Strict fuzz | 500 iterations, compatibility stub intentionally silent |
| Typecheck | 15/15 packages |
| RDE | 100/100 repos, 25,973 files each, target 48 → 0, no failures/additions |
| React Bench | 120/120 repos, 97,695 files each, target 425 → 0, no failures/additions |
| Oracle census | 3 tasks / 6 states, 6,478 files each, target 39 → 0, no failures/additions |

All 512 removed lane verdicts use the old synchronous-redraw message and classify as intended false positives; no true-positive contract remains under this rule ID.

## Test plan

- `nr --filter oxlint-plugin-react-doctor test`
- `FUZZ_RULE=no-cascading-set-state FUZZ_STRICT=1 FUZZ_ITERATIONS=500 nr --filter @react-doctor/fuzz fuzz`
- `nr typecheck`
- `nr lint`
- `nr format:check`
- `nr test`
- `nr smoke:json-report`

---

> [!NOTE]
> **Low Risk**
> Lint-only behavior change: removes a rule known to produce false positives; rule ID remains for configs but does not report by default.
> 
> **Overview**
> **Retires `react-doctor/no-cascading-set-state`** because counting synchronous `setState` calls inside one effect does not map to multiple React commits—React batches those updates into a single follow-up render, so the old warning and “use `useReducer`” guidance were misleading.
> 
> The rule implementation is replaced with a **default-off retired stub** via `defineRetiredRule` (the large setter-count / control-flow analyzer is removed). The rule ID stays registered for compatibility, but **`create()` no longer reports**.
> 
> **Test and harness updates** drop positive controls and expectations that depended on the old detector: liveness fixture removed, `KNOWN_UNCOVERED` documents retirement, verdict-robustness no longer audits this rule, integration tests expect silence on patterns that used to flag, and the massive regression suite collapses to “retired = no diagnostics.” A **fuzz regression seed** pins batched effect setters as a runtime-semantics case.
> 
> **Docs**: README example rule switched from `no-cascading-set-state` to `effect-needs-cleanup`; changeset notes the patch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e751d23c1f5ac4b90c341442ae1f2f00e8c7364c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rayhanadev (approved):
Adversarial review (4 lenses + refutation pass): 1 confirmed finding, 0 behavioral defects — the retirement stub is precedent-exact vs the two prior rn-* retirements and every behavioral change is pinned by a test that fails on main.
Fixed in b1a8126a1: removed dead CASCADING_SET_STATE_THRESHOLD from thresholds.ts (the retired detector was its only consumer).
Deslop pass found no safe simplification (81 added lines already minimal and precedent-matching).
Validation: oxlint-plugin-react-doctor suite green (549 files, 12262 tests passed) + tsc typecheck clean.
Follow-ups (not blocking): no-effect-chain.ts:36 prose still cross-references the retired rule as covering single-effect cascades — optional 2-line comment tweak.

PR #1201, opened 2026-07-13
fix(zod): recognize the official v4 export

aidenybai:
## Why

Catches Zod 4 migration issues when projects import the full classic API from the official `zod/v4` export instead of the package root.

Before:

```ts
import { z } from "zod/v4";

const email = z.string().email();
const schema = z.object({ value: z.string() }).strict();
```

After:

```ts
import { z } from "zod/v4";

const email = z.email();
const schema = z.strictObject({ value: z.string() });
```

## What changed

- Recognizes only the exact `zod` and `zod/v4` module sources across all four shared Zod migration-rule consumers.
- Keeps `zod/v3`, mini, core, locales, locale wildcards, internal-looking subpaths, and package lookalikes excluded.
- Covers named, renamed, namespace, default, and direct factory imports.
- Adds dedicated fuzz targets and generator snippets for all four affected rules.
- Keeps intentional liveness targets out of the confirmed-valid false-positive corpus oracle.

RDE was not run because cloud RDE does not exercise this AST-rule layer. The exact pinned local fixture was compared against its baseline: two diagnostics were added, both intended, with none removed.

## Test plan

- `nr --filter oxlint-plugin-react-doctor test src/plugin/rules/zod/zod-v4-no-deprecated-schema-apis.test.ts src/plugin/rules/zod/zod-v4-prefer-top-level-string-formats.regressions.test.ts src/plugin/rules/zod/zod-v4-no-deprecated-error-customization.test.ts src/plugin/rules/zod/zod-v4-no-deprecated-error-apis.test.ts`
- `FUZZ_RULE=zod-v4 FUZZ_STRICT=1 FUZZ_ITERATIONS=500 FUZZ_PRINT_SILENT=1 nr fuzz`
- `nr lint`
- `nr typecheck`
- `nr format:check`
- `nr smoke:json-report`

---

> [!NOTE]
> **Low Risk**
> Scoped change to Zod import-source matching and test/fuzz harness updates; no auth or runtime behavior outside lint rule coverage for two exact module strings.
> 
> **Overview**
> Zod v4 migration lint rules now treat imports from **`zod/v4`** the same as **`zod`**, so deprecated schema/error APIs and legacy string formats are reported when projects use the official v4 entry point.
> 
> Shared import resolution in `zod-ast.ts` uses `ZOD_MODULE_SOURCES` (`zod` and `zod/v4` only); subpaths like `zod/v4/core` and package lookalikes stay excluded. Four rules gain parameterized tests for named, namespace, default, and renamed imports from `zod/v4`, plus shared negative cases via `NON_FULL_ZOD_V4_MODULE_SOURCES`.
> 
> Fuzzing adds intentional **`corpus/targets`** programs for those rules, **`zod/v4`** snippets in generator pools, and narrows false-positive hunting to **`corpus/regressions`** so liveness targets are not treated as valid-code oracles.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ad02991ea4e5180b17c66d9bd25488e66be25518. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/tidy-zod-subpaths.md
+++ b/.changeset/tidy-zod-subpaths.md
@@ -1,0 +1,5 @@
+---
+"oxlint-plugin-react-doctor": patch
+---
+
+Recognize the official `zod/v4` export in Zod migration rules.
--- a/packages/fuzz/corpus/targets/README.md
+++ b/packages/fuzz/corpus/targets/README.md
@@ -1,0 +1,5 @@
+# Fuzz liveness targets
+
+These programs intentionally contain diagnostics. They keep changed rules on their reporting paths during fuzzing so strict runs exercise more than early exits.
+
+False-positive hunting loads only `corpus/regressions/`, where every program is valid by contract.
--- a/packages/fuzz/corpus/targets/zod-v4-no-deprecated-error-apis--official-v4-export.tsx
+++ b/packages/fuzz/corpus/targets/zod-v4-no-deprecated-error-apis--official-v4-export.tsx
@@ -1,0 +1,3 @@
+import { ZodError } from "zod/v4";
+
+export const flattenedError = new ZodError([]).flatten();
--- a/packages/fuzz/corpus/targets/zod-v4-no-deprecated-error-customization--official-v4-export.tsx
+++ b/packages/fuzz/corpus/targets/zod-v4-no-deprecated-error-customization--official-v4-export.tsx
@@ -1,0 +1,3 @@
+import { z } from "zod/v4";
+
+export const requiredSchema = z.string("Required");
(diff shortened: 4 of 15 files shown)


rayhanadev (approved):
Adversarial review (4 lenses, 2-round refutation): 0 proven / 0 confirmed findings — all 3 raised claims unanimously refuted (importKind blindness is pre-existing and only reachable from non-compiling TS; the two coverage claims are neutralized by always-on unit assertions and repo precedent).
Deslop: no safe simplification found (patch empty); the PR is already net-minimal (-11 lines removed vs 214 added, mostly test pins).
Re-validated at head e622865: 4 touched rule test files, 96/96 pass. Negative fixture defeats every plausible over-widening (zod/v3, zod/mini, zod/v4/*, @acme/zod/v4).
Follow-ups (not blocking): pre-existing `import { default as z }` false negative; pre-existing type-only-import importKind blindness (shared with "zod" on main); alias re-binding and dynamic import()/require() not followed (by design).

PR #1192, opened 2026-07-12
fix(rule): resolve forwardRef binding provenance

aidenybai:
## Problem

`forward-ref-uses-ref` inferred React semantics from the callee spelling alone. It reported unrelated local functions named `forwardRef` and missed the real React API when a named import was renamed.

## Change

Require the call to resolve to React's `forwardRef` export before applying callback arity guidance. The rule reuses the existing scope-aware React API resolver, including renamed named imports and immutable aliases. The shared resolver now also accepts statically computed members such as `React["forwardRef"]` while rejecting dynamic computed members.

Unbound globals, CommonJS calls, mutable aliases, dynamic properties, and callback identifiers remain conservative.

## Regression coverage

- unrelated local and other-package `forwardRef` helpers stay quiet
- direct, renamed, default, namespace, static-computed, optional, TypeScript-wrapped, and immutable multi-hop React aliases report
- mutable aliases, reassignment, lexical shadows, dynamic members, CommonJS, and unbound calls stay quiet
- zero-, two-, rest-parameter, and callback-identifier boundaries stay unchanged
- permanent fuzz seed covers the confirmed local-name collision false positive

## Validation

- focused rule/utility tests: 47 passed
- full oxlint plugin suite: 550 files passed; 12,320 tests passed; 148 skipped
- strict target fuzz: `FUZZ_RULE=forward-ref-uses-ref FUZZ_STRICT=1 FUZZ_ITERATIONS=500 nr fuzz`
- explicit fuzz fire coverage: 1/1 target rule fired
- fuzz corpus tests and false-positive hunt completed; no `forward-ref-uses-ref` finding on its valid seed
- performance probes remained near-linear through 2,000 direct calls and 800 multi-hop alias calls
- force-fresh RDE: 100/100 distinct pinned repositories, schema v3, no failures/partial scans; 15,047 -> 15,047 total diagnostics; target 2 -> 2; zero identity deltas; both retained target findings manually classified as true positives
- exact React Bench: 120/120 pinned repositories, no failures; 97,695 -> 97,695 extents; target 0 -> 0; zero added/removed findings; no solution-sensitive oracle shape found across 93 `forwardRef` files
- build, typecheck (15/15), lint, format check, JSON-report smoke, and `git diff --check` passed
- exact built-CLI reproduction reports the direct and renamed React violations and stays quiet on the local helper and two-parameter callbacks
- pre/post truffler and direct symbol searches found no duplicate helper; reused `isReactApiCall` and `getStaticPropertyName`
[...]

PR #1202, opened 2026-07-13
fix(prefer-use-effect-event): recognize stable callbacks

aidenybai:
## Why

`prefer-use-effect-event` treated every local `useCallback` dependency as identity-changing. A callback with an explicit empty dependency array, or dependencies proven to be stable React hook values, cannot make an effect re-subscribe on ordinary renders. React Bench surfaced the empty-array false positive on `write-react-sahil87-run-kit-282__7jJuPUp`.

## What changed

- Require proven React `useCallback` provenance before classifying a local callback dependency.
- Keep explicit empty dependency arrays quiet.
- Prove stable React hook dependencies such as refs, state setters, and reducer dispatch functions, including immutable multi-hop aliases.
- Preserve diagnostics for changing state, callback props, non-React hooks, nonempty unknown dependencies, dynamic lists, and omitted lists.
- Reject local shadows and similarly named APIs imported from userland packages.
- Add paired adversarial regression tests, permanent fuzz seeds, generator shapes for both valid and invalid paths, and a patch changeset.

## Validation

- Exact pinned Sahil model-patch replay, 156 files: baseline 3 target diagnostics, candidate 1. The two removed callbacks have `[]`; the retained `openComposeWithUploads` depends on a callback prop and remains a true positive.
- Focused rule regressions: 16 passed.
- Full plugin suite: 550 files passed, 12,323 tests passed, 148 skipped.
- Fuzz corpus: 37 passed, 400 skipped.
- `FUZZ_RULE=prefer-use-effect-event FUZZ_STRICT=1 FUZZ_ITERATIONS=500 nr fuzz`: passed; target fire coverage 1/1.
- Fresh focused RDE-100 at pinned revisions: 100/100 successful, baseline 69, candidate 54, 0 added, 15 removed. Manual classification: all 15 are stable `[]` callbacks and therefore removed false positives.
- Fresh focused React-Bench-120 at pinned revisions: 120/120 successful, 97,626 files per side, baseline 386, candidate 298, 0 added, 88 removed. Manual classification: 87 are stable `[]` callbacks; 1 depends only on two React `useRef` values. All 88 are removed false positives.
- `nr typecheck`: 15/15 tasks passed.
- `nr lint`: passed with existing repository warnings only.
- `nr format:check`
- `git diff --check`
- Direct GitHub review-thread query: 0 unresolved threads.

## Status

Leave as draft pending independent review. Do not merge.

---

> [!NOTE]
> **Low Risk**
> Lint-rule precision change with broad regression and fuzz coverage; behavior is more permissive (fewer warnings) but scoped to callback stability analysis, not runtime code paths.
> 
> **Overview**
> **`prefer-use-effect-event`** no longer flags every local **`useCallback`** in effect deps. It only treats bindings as identity-changing when they come from **proven React `useCallback`** and the dependency list is missing, non-static, or includes anything that is not a **stable React hook value** (empty `[]`, setters, refs, dispatch, `startTransition`, etc.), including **`const`** multi-hop aliases and **`React.useCallback`** / renamed imports.
> 
> Diagnostics stay for **`on[A-Z]`** callback props, changing state in deps, dynamic or omitted dep arrays, sparse holes, and **`let`**-aliased stable values. Non-React **`useCallback`** and shadows are ignored.
> 
> Adds a large regression suite, fuzz corpus seeds, and fuzz snippet shapes for stable vs changing callbacks; patch changeset for **`oxlint-plugin-react-doctor`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 755ce0a8205eb52a6f2dfca085cc7d9157522aa7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rayhanadev (approved):
Adversarial review (4 lenses → dedupe → 16 refuters → fixture proofs run against both head and main): 6 confirmed findings, 0 high/critical, verdict approve-with-nits.
All 6 fixed in cfd03546d: the `const { useCallback } = React` false negative (resolveNamedAliases at both isReactApiCall sites — collector and stability re-gate, so destructured stable deps don't FP), the vacuous stable-alias test converted to a changing-alias limitation pin, plus discriminating coverage pins for useRef / useTransition / useEffectEvent / useActionState deps, sparse-hole dep arrays (no crash), and the let-alias reject guard.
Merged main into the branch (a1e1ebaab) to pick up isDestructuredReactApiBinding from #1192, which the FN fix requires.
Deslop pass found no safe simplification (patch empty). Tests: oxlint plugin 12,796 passed / 554 files; eslint mirror, typecheck, lint, format:check all green.
Follow-ups (not blocking, pre-existing): rule still fires on stable-but-not-hook deps (literal/module consts, nested stable useCallback, userland stable wrappers, preact-renamed setters via getHookName's react-only source gate); alias bindings are never collected, so aliased changing callbacks stay silent (identical on main).

PR #1204, opened 2026-07-13
fix(rule): ignore async effect callback chains

aidenybai:
## Summary

- exclude async effect callbacks from `no-effect-chain` synchronous chain construction in both upstream-writer and downstream-reader positions
- preserve diagnostics for synchronous callback aliases and other synchronous near-neighbors
- add paired regression coverage, two permanent fuzz seeds, and verdict-preserving liveness coverage

Follow-up to #1182, which was merged before the async-callback review fix could land. Addresses https://github.com/millionco/react-doctor/pull/1182#discussion_r3568266314.

## Precision boundary

An async function executes synchronously until its first `await`, but passing an async callback directly to `useEffect` or `useLayoutEffect` violates React's effect callback contract and is separately reported by `no-async-effect-callback`. This rule therefore excludes the whole async effect from chain analysis instead of attempting partial await-position control flow. That prevents async effects from becoming either chain writers or chain readers; inline and named arrows, declarations, expressions, exact aliases, layout effects, and writes straddling `await` stay quiet. Synchronous exact-alias near-neighbors still report.

## Validation

- exact base/candidate A/B: async named upstream 1 to 0; async alias upstream 1 to 0; async inline and named downstream 1 to 0; synchronous alias 1 to 1
- focused `no-effect-chain` regressions: 44 passed
- strict fuzz: `FUZZ_RULE=no-effect-chain FUZZ_STRICT=1 FUZZ_ITERATIONS=500`; 1/1 target rule fired, 0 findings
- fuzz corpus and verdict robustness: passed; two permanent regression seeds included
- exact pinned local RDE A/B: 100/100 repositories, 25,973 files per build, baseline 63 and candidate 63 target findings, no added or removed diagnostics
- exact pinned React Bench A/B: 120/120 repositories, 97,626 files per build, baseline 221 and candidate 221 target findings, no added or removed diagnostics
- active React Bench `no-effect-chain` solution/oracle patches contain no async effect callback shape affected by this boundary
- oxlint plugin suite: 552 files, 12,628 passed, 148 skipped
- `nr typecheck`, `nr lint`, `nr format:check`, `nr smoke:json-report`, and `git diff --check`: passed
- full `nr test`: concurrent runs hit only known load-sensitive infrastructure tests (`spawn-batches`, two `performance-harness` timeouts, and one live TTY probe); the complete core package passed 1,294/1,294 in isolation, and the React Doctor performance/TTY tests pass in isolated runs

## Changeset

Patch changeset included for `oxlint-plugin-react-doctor`.

---

> [!NOTE]
> **Low Risk**
> Narrow precision fix in a lint rule with broad regression and fuzz coverage; no runtime or auth/data-path changes.
> 
> **Overview**
> **`no-effect-chain`** now skips **async** `useEffect` / `useLayoutEffect` callbacks when building cross-effect chains, so they are neither treated as upstream state writers nor downstream dep readers. The guard is `callback.async` at chain construction time (whole callback excluded, not partial control flow before `await`).
> 
> **Synchronous** chains are unchanged, including **exact local callback aliases** (e.g. `const effectCallback = loadSource`).
> 
> Adds regression tests for inline/named async upstream and downstream callbacks, layout effects, aliases, and writes straddling `await`; two fuzz corpus seeds; and **`no-effect-chain`** in verdict-preserving liveness coverage. Patch changeset for `oxlint-plugin-react-doctor`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6d337243bbdf7244bbf99dc493e467070d74e28a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/tidy-effect-callbacks.md
+++ b/.changeset/tidy-effect-callbacks.md
@@ -2,4 +2,4 @@
 "oxlint-plugin-react-doctor": patch
 ---
 
-Resolve exact local function bindings passed as effect callbacks.
+Resolve exact local function bindings passed as effect callbacks while excluding async callbacks from synchronous chain analysis.
--- a/packages/fuzz/corpus/regressions/no-effect-chain--async-downstream-effect-callback.tsx
+++ b/packages/fuzz/corpus/regressions/no-effect-chain--async-downstream-effect-callback.tsx
@@ -1,0 +1,21 @@
+// rule: no-effect-chain
+// weakness: control-flow
+// source: PR #1204 independent final audit
+
+import { useEffect, useState } from "react";
+
+export const AsyncDownstreamEffectCallback = ({ loadValue }) => {
+  const [source, setSource] = useState(0);
+  const [target, setTarget] = useState(0);
+
+  useEffect(() => {
+    setSource(1);
+  }, []);
+
+  const synchronizeTarget = async () => {
+    setTarget(await loadValue(source));
+  };
+
+  useEffect(synchronizeTarget, [loadValue, source]);
+  return target;
+};
--- a/packages/fuzz/corpus/regressions/no-effect-chain--named-async-effect-callback.tsx
+++ b/packages/fuzz/corpus/regressions/no-effect-chain--named-async-effect-callback.tsx
@@ -1,0 +1,22 @@
+// rule: no-effect-chain
+// weakness: control-flow
+// source: PR #1182 Bugbot review
+
+import { useEffect, useState } from "react";
+
+export const AsyncEffectCallback = ({ loadValue }) => {
+  const [source, setSource] = useState(0);
+  const [target, setTarget] = useState(0);
+
+  const loadSource = async () => {
+    await loadValue();
+    setSource(1);
+  };
+
+  useEffect(loadSource, [loadValue]);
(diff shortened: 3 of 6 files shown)


rayhanadev (approved):
Adversarial review: 1 proven finding, 0 further confirmed — all fixed in the pushed commit (6d337243b).
Fixed: the four "async downstream chain link" test variants were vacuous — their fetch call hit main's external-sync suppression, so they passed against the pre-fix rule; de-fetched to an opaque loadTargetValue loader mirroring the corpus fixture, and verified all four now fail against origin/main's rule and pass at head.
Deslop: no safe simplification found (empty patch, 0 lines). oxlint-plugin suite green on the rebased head: 553 files / 12,852 tests.

Follow-ups (not blocking):
- no-async-effect-callback cannot resolve identifier-passed callbacks (no context.scopes in its getEffectCallback call), so a named async effect callback with a pre-await synchronous write is now reported by no rule; passing context.scopes there would close the gap this PR's exemption widens.
- Sync generator effect callbacks are counted as writers (pre-existing FP, identical on main; no real-world occurrence).

PR #1225, opened 2026-07-13
fix(rule): prove cross-effect ref timer cleanup

aidenybai:
## Why

`effect-needs-cleanup` reported ReactTooltip's delay-change effect even though its ref-held timeout has both lifecycle guarantees: the effect clears the previous timeout before replacing it, and a stable mount-only effect clears the same timeout on unmount.

Before, the rule treated the companion `pendingOpenRef` early return as an unconditional path that skipped replacement cleanup. After, the rule accepts that path only when component-wide ref provenance proves that no timeout exists without the companion witness.

## What changed

- Prove negative companion-ref guards for ref-held timers across the component.
- Require every allocation to establish the same witness and release the previous handle.
- Require every witness clear to release the handle, or to null it after a proven timer-expiry callback consumes the witness.
- Reject ref escape, reassignment, mismatched timers, conditional cleanup, orphaning allocations, and release after the handle was nulled.
- Add paired regression coverage, a permanent fuzz seed, and a generated snippet-pool shape.
- Add a patch changeset for `oxlint-plugin-react-doctor`.

## Eval results

| Corpus | Repositories/files | Baseline | Candidate | Added | Removed |
| --- | ---: | ---: | ---: | ---: | ---: |
| RDE local oxlint fallback | 100 repositories | 159 | 159 | 0 | 0 |
| React Bench pinned scan | 120 repositories / 97,695 files per side | 436 | 436 | 0 | 0 |

Exact ReactTooltip task reconstruction:

- untouched source: 2 → 2
- model solution: 3 → 2, removing only the reported effect at line 935
- oracle solution: 2 → 2

## Test plan

- `nr test effect-needs-cleanup.regressions.test.ts` — 261 passed
- full `oxlint-plugin-react-doctor` suite — 13,111 passed, 148 skipped
- `FUZZ_RULE=effect-needs-cleanup FUZZ_STRICT=1 FUZZ_ITERATIONS=500 nr fuzz`
- fuzz corpus tests — 43 passed, 400 skipped
- `nr typecheck`
- `nr lint`
- `nr format:check`
- `nr smoke:json-report`
- `git diff --check`
- full repository suite: all affected packages and 2,241/2,243 React Doctor tests passed; the two live-stdin PTY probes fail identically on the untouched baseline in this local environment


aidenybai:
Independent adversarial review found this draft is not merge-ready despite its old green CI. The exact simplified pattern with rerun cleanup plus a separate unconditional unmount cleanup is already silent on current main. The remaining ReactTooltip model depends on a broader show-transition invariant that this local witness proof does not establish safely.

Concrete candidate regressions reproduced so far include inverse null-guard cleanup, setInterval treated as one-shot expiry, expiry callbacks that reschedule and lose the new handle, external/helper release followed by reallocation, and valid capture-before-null cleanup newly reported. I am holding the PR as draft and will not push/mark ready until the proof is conservative and the exact model plus all paired controls, strict fuzz, RDE, React Bench, CI, and review threads agree.

aidenybai:
Closing this implementation rather than weakening timer-leak detection. Current main already handles the simplified rerun-clear + separate unconditional unmount-cleanup pattern. The remaining ReactTooltip case is a real precision gap, but it depends on a broader cross-function show-transition invariant; the draft's witness-local proof reproduced multiple blocker false negatives and a cleanup false positive. The evidence stays recorded in the PR/ASAP audit for a future control-flow-grounded approach.

PR #1229, opened 2026-07-13
fix: no-impure-state-updater false positives on promise callbacks and event handlers

### LINKED ISSUE #1228: no-impure-state-updater: fires on promise callbacks and event handlers, not just state updaters (75 FPs / 77% of our errors)
### Summary

`no-impure-state-updater` fires on any single-argument arrow function that calls a `useState` setter, regardless of whether that function is a **state updater callback**. Promise callbacks and JSX event handlers are reported as "state updaters with side effects".

On our codebase (Next.js 15 app-dir, React 19, ~613 files) this produces **75 false positives across 49 files** — 77% of all error-severity findings, which drives the project score from a realistic number down to 29/100 ("Critical").

### Reproduction

Both of these are reported as `error  This state updater performs the nested state update "setX()"`:

```jsx
// 1. A promise callback — not an updater
const loadInvites = useCallback(() => {
  fetchInvites(api)
    .then((d) => setInvites(d))     // <-- reported
    .catch(() => setInvites([]))    // <-- reported
}, [api])

// 2. A JSX event handler — not an updater
<SelectMenu
  value={sort}
  onChange={(v) => setSort(v)}      // <-- reported
/>
```

Neither function is ever passed to a setter. The rule's actual target — a nested update inside an updater callback — looks like this, and is what we'd want flagged:

```jsx
setRows((prev) => {
  setTotal(prev.length)   // <-- THIS is the bug the rule is for
  return next
})
```

### Expected

The rule should only fire for functions passed **as the argument to a state setter** (`setX(fn)`), not for any callback that happens to call a setter in its body.

### Impact

Because the rule is error-severity and high-volume, it buries the real findings. In the same scan react-doctor correctly found genuine bugs we fixed (a `setTimeout` leaking past unmount, 18 refs mutated during render) — those were hard to see under 75 false errors. The tool's own output says "sample before you sweep… get the code owner's sign-off", which is good advice, but the sign-off here is "the code is correct".

### Environment

- react-doctor 0.7.7 (also reproduced on `@latest`)
- React 19, Next.js 15 (app dir), TypeScript

### Possibly related

Two other rules appear to flag correct code in the same scan (happy to file separately if useful):

- `effect-needs-cleanup` — still reports an effect **after** we added the missing cleanup; 3 of its 4 hits already had complete teardown (it seems unable to trace teardown targets through a nested helper closure, or past an early `return` guard).
- `no-locale-format-in-render` — still reports `Intl` formatters **after** they were hoisted to module scope; it now flags the `.format()` call site.

skoshx:
## Why

Fixes #1228: `no-impure-state-updater` fired on promise callbacks and event handlers — 77% of error-severity findings in the reporter's codebase (75 FPs across 49 files).

Before (all falsely reported as impure updaters):

```jsx
fetchInvites(api).then((d) => setInvites(d));
<select onChange={(e) => setSort(e.target.value)} />;
const handler = (newValue) => { setValue1(newValue); setValue2(newValue * 2); };
```

Root cause: for a reference whose binding is a **Parameter**, eslint-scope's `def.node` is the *enclosing function*, so `resolveToFunction` (`utils/effect/ast.ts`) returned that enclosing function and the rule analyzed it as if it were the updater callback.

## What changed

- `resolveToFunction` now bails for parameter-bound references via a new shared `hasParameterDef` helper — a parameter binding holds a value, never denotes its enclosing function.
- Since `resolveToFunction` is shared by five other consumers, each was audited for silent behavior changes:
  - `isCleanupReturnArgument` (`utils/effect/react.ts`): parameter-bound identifiers are now treated as conservatively cleanup-capable (`hasParameterDef(ref) || resolveToFunction(ref)`), preserving the old "unknown callable counts as cleanup" behavior for `return cleanup` where `cleanup` is a hook/component parameter.
  - `getEffectFn`, `resolvesToAsyncFunction`/`isSyncStateSetterCall`, `findStateSetterReference`, `no-pass-live-state-to-parent`, `no-pass-data-to-parent` handler-bag/argument paths: verified safe with fixtures against `main`; no behavior regressions.
- Dedup: removed the local `hasParameterDef`/`resolvesToFunctionBinding` in `no-pass-data-to-parent` and the redundant `Parameter` half of the guard in `collect-effect-state-write-facts` (its `ImportBinding` guard remains) — the check is now centralized in `resolveToFunction`.
- Regression tests added across `no-impure-state-updater`, `no-event-handler`, `no-effect-event-handler`, `no-chain-state-updates`, `no-effect-chain`, `no-pass-data-to-parent`, `no-pass-live-state-to-parent`, `prefer-use-sync-external-store`.

## Eval results

RDE was not run (fix narrows detection; no new detection surface). Fuzzing instead, all `FUZZ_STRICT=1 FUZZ_ITERATIONS=500` with confirmed 1/1 fire coverage: `no-impure-state-updater`, `effect-needs-cleanup`, `no-pass-data-to-parent`, `no-pass-live-state-to-parent` — all passed.

## Test plan

- `pnpm vitest run` on the state-and-effects suite (61 files, 1546 tests pass)
- `pnpm typecheck`
- `pnpm lint`, `pnpm format:check`

Link to Devin session: https://app.devin.ai/sessions/88bf7b1a41e14488ba5e01671cdd0fb4
Requested by: @aidenybai

---

> [!NOTE]
> **Medium Risk**
> Touches a shared AST helper used by multiple effect/state rules; behavior was audited and fuzzed, but incorrect resolution could still shift diagnostics in other rules.
> 
> **Overview**
> Fixes **`no-impure-state-updater`** reporting promise `.then` handlers, `onChange` callbacks, and similar patterns as impure state updaters when they only call `setState`.
> 
> **`resolveToFunction`** (`utils/effect/ast.ts`) now returns early for references bound to a **Parameter** via a shared **`hasParameterDef`** helper, because eslint-scope’s `def.node` for parameters points at the enclosing function and previously made the rule treat that outer function as the updater callback.
> 
> Because **`resolveToFunction`** is shared, related paths were checked or adjusted: **`isCleanupReturnArgument`** treats parameter-bound identifiers as conservatively cleanup-capable so hook/component `return cleanup` behavior stays the same; duplicate parameter checks in **`no-pass-data-to-parent`** and **`collect-effect-state-write-facts`** were removed in favor of the centralized guard.
> 
> Regression coverage was added across several state/effects rules; fuzz runs passed with no reported behavior regressions on audited consumers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e75d9f11a79480debb6a4614ecafdb8938937be0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aidenybai:
Closing as superseded by #1249 (`4fce1c1bf`). The current branch has an empty net diff from its merge base, while main already contains the shared `hasParameterDefinition` guard, cleanup-preservation behavior, paired cross-rule regressions, and the changeset. Rebasing or merging this PR would add no code.

PR #1232, opened 2026-07-13
fix(effect-needs-cleanup): prove conditional timer cleanup

aidenybai:
## Summary

- recognize timers allocated in a value-preserving conditional expression and stored in an immutable local handle
- accept cleanup only when the returned effect cleanup releases that exact handle on every path
- preserve diagnostics for mutable, aliased, transformed, mismatched, or partially cleaned timer handles
- add paired rule regressions plus permanent fuzz regression and target seeds

## Grounding

React Bench job `fix-react-rdh-lewhunt-divz-divz__yioQpdq` pins `lewhunt/divz` at `9e44f95806f940df9f21273176efa552ed09b6c4`.

The verifier solution conditionally assigns `setInterval(...)` to `autoplayTimer` and returns a cleanup that guards and clears that same handle. Runtime reproduction confirms each allocated interval is cleared and the final active interval count is zero. The job tests pass, but stored React Doctor 0.7.6 reports `effect-needs-cleanup`, making `rd_ok=0`; published 0.7.2 and 0.7.4 do not reproduce this specific regression.

Exact-source target diagnostics:

| Source | `origin/main` | This PR |
| --- | ---: | ---: |
| untouched pinned source | 0 | 0 |
| verifier/model patch | 1 | 0 |
| oracle patch | 0 | 0 |

## Precision boundary

The detector follows the resource call only through transparent wrappers, a selected conditional arm, and the final position of a sequence expression into a stable `const` binding. For conditional timers, the returned cleanup must release the exact binding symbol, and the release or its exact-handle guard must cover every cleanup path.

Opaque calls, construction, arithmetic transforms, non-final sequence values, aliases, mutable assignments, wrong handles/APIs, missing cleanup, unrelated guards, and early exits remain diagnostics.

## Validation

- focused and full plugin tests: 554 files; 13,113 passed, 148 skipped
- strict fuzz: `FUZZ_RULE=effect-needs-cleanup FUZZ_STRICT=1 FUZZ_ITERATIONS=500 nr fuzz`; 354 executions, 95 target-rule firings, 0 findings
- React Bench-derived fuzz corpus: 27 of 32 target files copied, 0 missing repositories, strict 500 clean
- isolated local RDE: independent frozen installs and builds for baseline/candidate; source maps match each checked-out rule source; 100/100 pinned roots scanned across 12 repositories on both sides, 0 scan failures, target diagnostics 42 to 42, no added or removed verdicts
- exact-head GitHub CI matrix: all checks passed on Linux, macOS, and Windows across Node 20, 22, 24, 25, and 26
- `nr typecheck`
- `nr lint`
- `nr format:check`
- `nr smoke:json-report`
- `git diff --check`

### DIFF
--- a/.changeset/dark-parks-double.md
+++ b/.changeset/dark-parks-double.md
@@ -1,0 +1,5 @@
+---
+"oxlint-plugin-react-doctor": patch
+---
+
+Fix `effect-needs-cleanup` false positives for conditionally allocated resources whose exact stable handle is released by the returned effect cleanup.
--- a/packages/fuzz/corpus/regressions/effect-needs-cleanup--conditional-timer-handle.tsx
+++ b/packages/fuzz/corpus/regressions/effect-needs-cleanup--conditional-timer-handle.tsx
@@ -1,0 +1,14 @@
+// rule: effect-needs-cleanup
+// weakness: control-flow
+// source: React Bench Divz trial fix-react-rdh-lewhunt-divz-divz__yioQpdq
+import { useEffect } from "react";
+
+export const Autoplay = ({ playing }: { playing: boolean }) => {
+  useEffect(() => {
+    const timerId = playing ? setInterval(() => advance(), 1000) : undefined;
+    return () => {
+      if (timerId) clearInterval(timerId);
+    };
+  }, [playing]);
+  return null;
+};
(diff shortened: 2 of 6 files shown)


aidenybai:
Closing as obsolete after re-grounding on the current `origin/main` (`9fc93f865c7ef2638bfa802f0544256fd5fc85e5`).

I discarded the earlier comparison because a cross-worktree dependency link made that evidence untrustworthy, then rebuilt current main from a fresh independent clone and replayed the exact pinned `lewhunt/divz` SHA `9e44f95806f940df9f21273176efa552ed09b6c4`.

Fresh complete schema-v3 scans now show:

- untouched pinned source: target 0
- exact verifier/model patch: target 0, 36 total diagnostics
- exact oracle patch: target 0, 39 total diagnostics
- this PR: the same target 0 and identical diagnostic identities for both patched states

Current main therefore no longer reproduces the claimed false positive, and this branch has no real-repository verdict to fix. Keeping its additional detector surface would be speculative, so it should not land.

PR #1240, opened 2026-07-13
fix(effect-needs-cleanup): prove pre-allocation timer release

aidenybai:
## Why

`effect-needs-cleanup` rejects a ref-held timeout whose previous handle is cleared before every replacement and whose current handle is cleared by a separate unmount effect. The false positive occurs when non-allocating early returns appear before or after the rerun clear.

This draft starts with the exact failing regression. The detector change and adversarial controls will follow before review.

## Current MVP

- Adds the helper-mediated split-cleanup regression grounded in the ReactTooltip timing trial.
- Confirms current `origin/main` reports one diagnostic where zero are expected.

## Required before review

- Prove release dominance only on paths reaching the allocation.
- Add allocation, reassignment, reschedule, mismatched-ref, and orphan controls.
- Run focused/full tests, strict fuzz, RDE, React Bench parity, and review-thread audit.

Do not merge automatically.

### DIFF
--- a/packages/oxlint-plugin-react-doctor/src/plugin/rules/state-and-effects/effect-needs-cleanup.regressions.test.ts
+++ b/packages/oxlint-plugin-react-doctor/src/plugin/rules/state-and-effects/effect-needs-cleanup.regressions.test.ts
@@ -2685,6 +2685,42 @@ export const Debounced = ({ value }) => {
     expect(result.diagnostics).toHaveLength(0);
   });
 
+  it("accepts helper-mediated timer cleanup after non-allocating early returns", () => {
+    const result = runRule(
+      effectNeedsCleanup,
+      `import { useEffect, useRef } from "react";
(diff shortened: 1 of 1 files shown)


aidenybai:
Closing after adversarial validation. The broad split-lifecycle implementation introduced real false negatives around orphaned handles (conditional/double/async/helper-based replacements, completion callbacks reused elsewhere, intervals, and retained sibling schedulers). Narrowing the ownership proof fixed those regressions, but the authentic ReactTooltip source still remains at the baseline 3 diagnostics because its safety depends on a larger token/state-machine invariant across sibling schedulers and commit call sites. The resulting ~1,200-line change no longer fixes the reported case or qualifies as an MVP, so it should not ship under this claim. A future fix needs a separately researched one-shot revocation contract and matching diagnostic wording, or the source should keep cleanup local.

PR #1247, opened 2026-07-14
fix(effect-needs-cleanup): own promise-callback timers behind an active flag

### LINKED ISSUE #1241: [effect-needs-cleanup] false positive when timeout is created in a promise callback
## Summary

`effect-needs-cleanup` reports an effect even though every timer it can create is owned and released by the effect cleanup. The timer is created inside a Promise callback, so the detector sees the nested `setTimeout` but does not associate it with the returned teardown.

## Reproduction

```tsx
useEffect(() => {
  let isActive = true
  let permissionRecheckTimeout: ReturnType<typeof setTimeout> | undefined

  void syncReminder().then((result) => {
    if (!isActive || result !== 'permission-pending') return

    permissionRecheckTimeout = setTimeout(() => {
      if (isActive) setPermissionCheckTick((tick) => tick + 1)
    }, 5_000)
  })

  return () => {
    isActive = false
    if (permissionRecheckTimeout) clearTimeout(permissionRecheckTimeout)
  }
}, [syncReminder])
```

The effect owns the timer in both relevant cases:

- If the Promise resolves after unmount, `isActive` is false and no timer is created.
- If the timer was created before unmount, the returned cleanup calls `clearTimeout` on its handle.

React Doctor 0.7.7 reports `react-doctor/effect-needs-cleanup` on the `useEffect` despite this teardown.

## Expected behavior

Do not report the effect when a timer created by a nested Promise callback is guarded by an active flag and its handle is cleared by the effect's returned cleanup. The rule should trace the timer handle captured by the nested callback to the cleanup function, or conservatively recognize this ownership pattern.

## Environment

- react-doctor 0.7.7 (`@latest`)
- React 19.2.7
- Expo ~55 / React Native
- Command: `npx react-doctor@latest --verbose --scope changed`

## Related

- #306 covers a different cleanup-tracking false positive involving `AbortController`.
- #1228 mentions `effect-needs-cleanup` as a possibly related false-positive family, but focuses on `no-impure-state-updater`.

## Validation note

In the same scan, a second `effect-needs-cleanup` finding on a polling hook was not classified as a false positive: its cleanup clears the current timer but does not invalidate an in-flight poll that can schedule another timer after unmount.

rayhanadev:
## Why

Fixes #1241. `effect-needs-cleanup` flagged a `useEffect` even though every timer it can create is owned and released by the returned cleanup. A `setTimeout` created inside a Promise `.then` continuation lives in a different control-flow graph than the effect's cleanup `return`, so the rule's path-coverage check can never connect the two and the effect is reported despite proper teardown.

Before:

```tsx
useEffect(() => {
  let isActive = true
  let permissionRecheckTimeout
  void syncReminder().then((result) => {
    if (!isActive || result !== 'permission-pending') return
    permissionRecheckTimeout = setTimeout(() => {
      if (isActive) setPermissionCheckTick((tick) => tick + 1)
    }, 5_000)
  })
  return () => {
    isActive = false
    if (permissionRecheckTimeout) clearTimeout(permissionRecheckTimeout)
  }
}, [syncReminder])
// react-doctor/effect-needs-cleanup  ← false positive
```

After:

```
(no diagnostic — the timer is owned)
```

## What changed

- `effectHasCleanupForUsage`: when the standard CFG path check fails, fall back to `hasGuardedDeferredTimerOwnership`. A timer whose handle is stored in an effect-scope variable is recognized as owned when the returned cleanup **both** clears that handle **and** assigns a scalar-falsy value to an "active" flag (a `let`/`var` or `ref.current`, key-matched via `resolveExpressionKey`) that **structurally gates** the timer's creation — an enclosing `if (flag)` or a preceding `if (!flag …) return`.
- Soundness rests on the guard: a continuation that resolves after unmount reads the falsy flag and never schedules; one that resolved before unmount stored a handle the cleanup clears.
- The issue's own validation note is respected: a self-rescheduling poll whose cleanup clears the current timer but never invalidates a flag **still fires** (no guard → not exempt). Added regression tests pin this, plus the "handle cleared but flag alive" and "timer created before the guard read" leak variants.
- Reuses the shared `statementAlwaysExits` util for the early-return guard instead of a bespoke check.
- Added a changeset (`oxlint-plugin-react-doctor`, patch). The ESLint plugin re-exports this rule, so the fix propagates automatically.

This supersedes the automated triage PR #1244, which exempts on "handle cleared" alone and would turn the polling true-positive into a false negative.

## Test plan

- `pnpm --filter oxlint-plugin-react-doctor test` — 13,101 passed / 148 skipped, incl. 6 new `effect-needs-cleanup` regression tests (3 FP-fixed: reported case, `if (isMounted)` block, ref guard; 3 TP-kept: polling, flag-not-invalidated, timer-before-guard).
- `pnpm --filter oxlint-plugin-react-doctor typecheck` — clean.
- `vp lint` / `vp fmt --check` on touched files — clean.
- Built plugin + `packages/react-doctor/tests/regressions/state-rules/effect-needs-cleanup.test.ts` e2e — 37 passed.

## Follow-ups (not in this PR, to keep it scoped)

- A timer in a **synchronous anonymous IIFE** inside an effect (`(() => { timer = setTimeout(...) })(); return () => clearTimeout(timer)`) is still a false positive — same anchor-lift gap, but no guard flag to rescue it. The clean fix is lifting the cleanup path anchor to the IIFE call site for synchronously-invoked anonymous functions.
- `no-fetch-in-effect.ts` has a parallel cancellation-flag guard walker (`isCompletionSinkGuardedByCancellationFlag`). A future refactor could extract a shared, polarity-parametric guard util; kept separate here to avoid touching a second rule's behavior in a targeted fix.

---

> [!NOTE]
> **Medium Risk**
> Adds non-trivial static analysis to a widely used lint rule; behavior is constrained by regression tests but guard detection could miss or over-accept edge cases.
> 
> **Overview**
> Fixes **#1241** false positives where `effect-needs-cleanup` reported `useEffect` timers scheduled inside Promise `.then` handlers even when teardown was correct—the rule’s CFG path check cannot link deferred timer creation to the effect’s cleanup `return`.
> 
> **`effectHasCleanupForUsage`** now falls back to **`hasGuardedDeferredTimerOwnership`** when path coverage fails. Timers created in a nested callback (e.g. Promise continuation) count as owned only if cleanup **clears the stored handle** and **sets an effect-scope “active” flag to a scalar-falsy value** (`isActive`, `isMounted`, `ref.current`, etc.), and **`setTimeout` is structurally gated** by that flag (`if (!isActive) return`, `if (isMounted) { … }`, or similar). Self-rescheduling polls and cases that clear the timer without invalidating the guard still diagnose.
> 
> Adds six regression tests and a patch **changeset** for `oxlint-plugin-react-doctor`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 54262e91c0864f4e7952de5992fd4931ff60549f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/effect-cleanup-guarded-promise-timer.md
+++ b/.changeset/effect-cleanup-guarded-promise-timer.md
@@ -1,0 +1,5 @@
+---
+"oxlint-plugin-react-doctor": patch
+---
+
+Stop effect-needs-cleanup from flagging a timer created inside a Promise callback when an effect-scope active flag gates its creation and the returned cleanup both clears the handle and invalidates that flag; a self-rescheduling poll with no such flag still fires.
--- a/packages/oxlint-plugin-react-doctor/src/plugin/rules/state-and-effects/effect-needs-cleanup.regressions.test.ts
+++ b/packages/oxlint-plugin-react-doctor/src/plugin/rules/state-and-effects/effect-needs-cleanup.regressions.test.ts
@@ -4066,3 +4066,150 @@ export const LanguageProvider = ({ i18next }) => {
     expect(result.diagnostics).toHaveLength(1);
   });
 });
+
+describe("effect-needs-cleanup regressions (issue #1241 promise-callback timer with active-flag guard)", () => {
+  it("does not flag a promise-callback timer guarded by an isActive early-return with its handle cleared", () => {
+    const result = runRule(
+      effectNeedsCleanup,
+      `import { useEffect, useState } from "react";
+export const PermissionWatcher = ({ syncReminder }) => {
+  const [, setPermissionCheckTick] = useState(0);
+  useEffect(() => {
+    let isActive = true;
+    let permissionRecheckTimeout;
+    void syncReminder().then((result) => {
+      if (!isActive || result !== "permission-pending") return;
+      permissionRecheckTimeout = setTimeout(() => {
+        if (isActive) setPermissionCheckTick((tick) => tick + 1);
+      }, 5000);
+    });
+    return () => {
+      isActive = false;
+      if (permissionRecheckTimeout) clearTimeout(permissionRecheckTimeout);
+    };
+  }, [syncReminder]);
+  return null;
+};`,
+    );
+    expect(result.parseErrors).toEqual([]);
+    expect(result.diagnostics).toHaveLength(0);
+  });
+
+  it("does not flag a promise-callback timer wrapped in an `if (isMounted)` block with its handle cleared", () => {
+    const result = runRule(
+      effectNeedsCleanup,
+      `import { useEffect } from "react";
+export const Retrier = ({ load }) => {
+  useEffect(() => {
+    let isMounted = true;
+    let retryTimer;
+    void load().then(() => {
+      if (isMounted) {
+        retryTimer = setTimeout(() => {}, 1000);
+      }
+    });
+    return () => {
+      isMounted = false;
+      clearTimeout(retryTimer);
+    };
+  }, [load]);
+  return null;
+};`,
+    );
+    expect(result.parseErrors).toEqual([]);
+    expect(result.diagnostics).toHaveLength(0);
+  });
+
+  it("does not flag a promise-callback timer guarded by a ref invalidated in cleanup", () => {
+    const result = runRule(
+      effectNeedsCleanup,
+      `import { useEffect, useRef } from "react";
+export const Retrier = ({ load }) => {
+  const mountedRef = useRef(true);
+  useEffect(() => {
+    let timer;
+    void load().then(() => {
+      if (!mountedRef.current) return;
+      timer = setTimeout(() => {}, 1000);
+    });
+    return () => {
+      mountedRef.current = false;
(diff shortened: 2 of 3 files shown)


aidenybai:
Closing as superseded by #1250, merged in `c79c897ec6128a3155ae23d10dda3f77e28683a5`.

Current main now contains the exact #1241 Promise-callback timer regression family, paired leak controls, a permanent fuzz seed, and a broader ownership proof that also tracks the Promise callback through `collectEffectInvokedFunctions`. The exact active-flag/handle pattern is clean on main, while missing invalidation, pre-guard scheduling, and self-rescheduling controls remain diagnostics.

This PR now conflicts with that merged implementation and would duplicate the same detector surface. There is no remaining current-main verdict for it to fix, so rebasing or merging it would add redundant competing logic rather than resolve a live issue.

PR #1251, opened 2026-07-14
fix(rules-of-hooks): preserve typed Hook adapter render provenance

aidenybai:
## Problem

At alovajs/alova@75e3f473b24c73f6fc44fa0757d137fac38f2bad, `packages/client/src/statesHook/react.ts` exports the React implementation of alova’s typed `StatesHook` adapter. React Doctor currently reports 12 `rules-of-hooks` errors because its object methods are named `create`, `memorize`, `ref`, `effectRequest`, `computed`, `watch`, `onMounted`, and `onUnmounted`.

These methods are reached synchronously by alova custom Hooks through `statesHookHelper`; React runtime Hook identity depends on stable render execution, not on the adapter property spelling.

## Draft state

This first draft intentionally contains the exact pinned regression only. It currently fails with all 12 reproduced diagnostics. The detector change and adversarial paired controls will follow after the proof boundary is implemented.

## Precision boundary

The fix must require statically resolved render-flow provenance. It must not exempt arbitrary object callbacks or use property/type names as an allowlist. Conditional, looped, event-time, escaping, and opaque calls must continue to report.

## Validation plan

- paired direct/member/alias/wrapper/conditional/event controls
- permanent fuzz corpus seed and generator coverage
- strict 500-iteration target-rule fuzz run with target firing
- focused and full package checks
- current-vs-candidate RDE and pinned React Bench/oracle A/B
- terminal CI and unresolved review-thread audit

### DIFF
--- a/packages/oxlint-plugin-react-doctor/src/plugin/rules/react-builtins/rules-of-hooks-adapters.regressions.test.ts
+++ b/packages/oxlint-plugin-react-doctor/src/plugin/rules/react-builtins/rules-of-hooks-adapters.regressions.test.ts
@@ -1,0 +1,83 @@
+import { describe, expect, it } from "vite-plus/test";
+import { runRule } from "../../../test-utils/run-rule.js";
+import { rulesOfHooks } from "./rules-of-hooks.js";
+
+describe("react-builtins/rules-of-hooks — typed Hook adapters", () => {
+  it("accepts the pinned alova React StatesHook adapter", () => {
+    const result = runRule(
+      rulesOfHooks,
+      `
+        import { falseValue, isNumber, noop, undefinedValue } from "@alova/shared";
+        import { StatesHook } from "alova";
+        import { MutableRefObject, useCallback, useEffect, useMemo, useRef, useState } from "react";
+        import { ReactHookExportType, ReactState } from "~/typings/stateshook/react";
+
+        const stateToData = <D>(reactState: ReactState<D>) => (2 in reactState ? reactState[2] : reactState[0]);
+        const refCurrent = <T>(ref: MutableRefObject<T>) => ref.current;
+        const setRef = <T>(ref: MutableRefObject<T>, newVal: T) => {
+          ref.current = newVal;
+        };
+
+        export default {
+          name: "React",
+          create: initialValue => useState(initialValue),
+          export: stateToData,
+          dehydrate: stateToData,
+          update: (newVal, state) => {
+            state[2] = newVal;
+            state[1](newVal);
+          },
+          memorize: fn => {
+            const fnRef = useRef(noop as typeof fn);
+            setRef(fnRef, fn);
+            return useCallback((...args: any[]) => refCurrent(fnRef)(...args), []);
+          },
+          ref: initialValue => {
+            const refObj = useRef(initialValue);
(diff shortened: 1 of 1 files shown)


aidenybai:
Closing after exact semantic validation; this is not safe to fix with a detector exemption.

Evidence:
- Current React Doctor and upstream `eslint-plugin-react-hooks@5.0.0` both emit the same 12 `rules-of-hooks` diagnostics at the same locations.
- React documents that custom Hook names must start with `use`, specifically so stateful logic cannot hide behind ordinary calls. These adapter property functions are named `create`, `memorize`, `ref`, `effectRequest`, `computed`, `watch`, `onMounted`, and `onUnmounted`.
- Owned source has 76 direct calls through `statesHookHelper` wrappers (create 22, ref 36, computed 4, effectRequest 1, memorize 3, watch 3, onMounted 3, onUnmounted 4); those observed calls are direct in their current owners. The pinned build passes, as do `usePagination` (40/40), action delegation (2/2), and auto request (5/5). That confirms the tested paths work, not that the adapter methods are universally safe.
- The same contract is dynamically framework-selected through mutable global `boundStatesHook`; `createAlova` can rebind the object when the adapter name matches. Public `actionDelegationMiddleware` and `statesHookHelper` create unresolved downstream call paths.
- `exposeProvider` calls the React adapter’s `memorize` wrapper from a `for...in` branch based on runtime property/value/tracking state, so stable Hook count cannot be proven from the adapter file or its type assertion.

A type/property-name exception would hide real conditional/event/loop misuse. A whole-program proof cannot be complete for this exported runtime-selected API. The correct code-side way to make these custom Hook boundaries statically valid is `use`-named wrappers (or an explicit source suppression owned by alova), not weakening React Doctor’s definite direct-violation rule. This draft contained only a deliberately failing reproduction, so it is being closed rather than left as a pseudo-fix.

aidenybai:
Current-main re-audit confirms the closure decision.

- Replayed the exact pinned adapter on `origin/main` at `6e5c240b610eb7c020d05200aec173aed162547e`: React Doctor emits the same 12 `rules-of-hooks` diagnostics.
- `eslint-plugin-react-hooks@7.1.1` independently reports the same object-property callback shape.
- The authentic alova checkout at `75e3f473b24c73f6fc44fa0757d137fac38f2bad` builds successfully, and its owned `usePagination`, action-delegation, and auto-request suites pass 47/47. Those tests establish current happy-path behavior, but not universal Hook order.
- The adapter is a public default export. `createAlova` stores the selected adapter in mutable module-global `boundStatesHook`, `promiseStatesHook()` returns that runtime-selected object, and `statesHookHelper()` forwards its methods. `exposeProvider()` also invokes the `memorize` method from a runtime `for...in` branch.
- A React 18 runtime replay using the authentic exported `alova/react` adapter renders `reactStatesHook.create(0)` on one render and skips it on the next. React reports the changed Hook order and throws `Rendered fewer hooks than expected`.

A detector exception based on `StatesHook`, `as`/`satisfies`, or object-property names would therefore hide reachable conditional/event/loop misuse. The per-file rule cannot prove every consumer of this exported runtime-selected object. No detector change, changeset, fuzz seed, RDE run, or Bench A/B is warranted because this candidate does not satisfy the proposed "all resolved calls are stable render-flow calls" boundary.

PR #1261, opened 2026-07-14
fix(rule): ignore proven analytics-only interactions

aidenybai:
## Why

Nexu open-design at `52611c07dbf922b775d813d6af1323c40fe20a52` attaches an `onClick` to `IntegrationsView.tsx:170` solely to emit the `integrations / skills_tab / coming_soon` analytics event. The section exposes no navigation, state change, submission, selection, or other user-facing operation. Runtime replay records the analytics event while leaving the DOM unchanged, so keyboard activation is not an equivalent product action that the section must expose.

Baseline `ece35b3ff` reports one `no-noninteractive-element-interactions` diagnostic at that exact site. This branch reports zero there.

## What changed

- Ignores an interaction handler only when every analyzed path is observation-only and at least one observation call has verified SDK provenance.
- Follows stable local wrappers and bounded named/namespace cross-file calls used by the authentic Nexu chain.
- Keeps unknown callbacks, unverified analytics-named factories, state/navigation, assignments, opaque member/destructuring reads, mixed branches, and function callbacks passed to analytics methods reportable.
- Declares the rule cross-file and fingerprints its actually called imports for sidecar-cache correctness.
- Adds a patch changeset and permanent fuzz corpus seed.

## Precision boundary

This does not infer safety from names such as `track` or from an `analytics/` path alone. A local factory must lead to a recognized analytics SDK import. Getter/proxy-capable reads and callback arguments remain conservative unless the value is proven plain local data. Analysis is depth-bounded; unresolved or deeper call graphs continue to report.

## Tests

- Exact pinned Nexu source: target diagnostic `1 -> 0`, parse-clean.
- Paired positives cover state, navigation, opaque callbacks, reassignment, dynamic methods, mixed branches, switch-case effects, unsafe defaults, unverified factories, callback invocation, and getter-backed payload/destructuring reads.
- Clean counterexamples cover direct PostHog capture, stable aliases, namespace/optional calls, supplied arguments that skip defaults, switch `break`, and stable computed analytics methods.
- Focused cross-file rule tests: 24 passed.
- Full oxlint plugin: 560 files, 13,988 passed, 181 skipped.
- Full core: 118 files, 1,376 passed.
- Plugin typecheck and cross-file registry tests pass.
- Strict fuzz: `FUZZ_RULE=no-noninteractive-element-interactions FUZZ_STRICT=1 FUZZ_ITERATIONS=500 nr fuzz` passes; direct stats show 40 fired programs / 333 executed, 221 parse skips, zero findings.

The PR remains draft until the replacement CI matrix and review audit are both clean.

### DIFF
--- a/.changeset/afraid-regions-look.md
+++ b/.changeset/afraid-regions-look.md
@@ -1,0 +1,5 @@
+---
+"oxlint-plugin-react-doctor": patch
+---
+
+Avoid no-noninteractive-element-interactions false positives for handlers proven to perform analytics-only observation.
--- a/packages/core/tests/cross-file-rule-ids.test.ts
+++ b/packages/core/tests/cross-file-rule-ids.test.ts
@@ -154,6 +154,7 @@ describe("CROSS_FILE_RULE_IDS", () => {
       "no-locale-format-in-render",
       "no-match-media-in-state-initializer",
       "no-mutating-reducer-state",
+      "no-noninteractive-element-interactions",
       "no-unguarded-browser-global-in-render-or-hook-init",
       "only-export-components",
       "prefer-dynamic-import",
--- a/packages/fuzz/corpus/regressions/no-noninteractive-element-interactions--analytics-only-handler.tsx
+++ b/packages/fuzz/corpus/regressions/no-noninteractive-element-interactions--analytics-only-handler.tsx
@@ -1,0 +1,16 @@
+// rule: no-noninteractive-element-interactions
+// weakness: cross-file
+// source: Nexu open-design 52611c07, IntegrationsView.tsx:170
+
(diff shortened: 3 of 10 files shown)


aidenybai:
Closing as architecturally unsafe. The authentic FP is real, but the proposed exemption cannot soundly prove the returned analytics capability: Nexu's track flows through React context/provider values, useMemo/useCallback, async initialization, fetch wrapping, and a lazy SDK client. Independent adversarial review reproduced false negatives from unused SDK imports and callback-capable wrappers. Name/path/import provenance is insufficient, and a correct fix requires a dedicated return-capability/effect dataflow design before implementation. No code from this draft should merge.

PR #1275, opened 2026-07-14
fix(no-interactive-element-to-noninteractive-role): allow native td cell role

aidenybai:
## Why

The pinned `hyparam/hightable@6ab8470aaae96680cb5db02ce3625af19dac91cc` benchmark uses an interactive native `<td role="cell">` at `src/components/Cell/Cell.tsx:108`. React Doctor reports that the role removes the element's interactive semantics, but `cell` is the native-equivalent ARIA role of a table data cell and does not erase its focus or keyboard handlers.

The boundary remains narrow: a genuinely mismatched interactive element such as `<button role="cell">` must still report.

## Current draft state

- Adds the exact minimized Hightable shape as a regression fixture.
- Adds a paired `<button role="cell">` true-positive control.
- The valid regression intentionally fails on this draft commit.

Implementation, permanent fuzz coverage, pinned source/oracle comparisons, and full validation will follow in subsequent commits.

## Evidence

- React Bench job: `write-react-hyparam-hightable-45__fZG7bpo`
- Stored diagnostic: `src/components/Cell/Cell.tsx:108:7`
- W3C ARIA in HTML: `<td>` exposes `cell` in a semantic table; explicitly restating the implicit role is allowed though not recommended.
- WAI-ARIA: `cell` has HTML `<td>` as its base concept.

### DIFF
--- a/packages/oxlint-plugin-react-doctor/src/plugin/rules/a11y/no-interactive-element-to-noninteractive-role.regressions.test.ts
+++ b/packages/oxlint-plugin-react-doctor/src/plugin/rules/a11y/no-interactive-element-to-noninteractive-role.regressions.test.ts
@@ -1,0 +1,23 @@
+import { describe, expect, it } from "vite-plus/test";
+import { runRule } from "../../../test-utils/run-rule.js";
+import { noInteractiveElementToNoninteractiveRole } from "./no-interactive-element-to-noninteractive-role.js";
+
+describe("a11y/no-interactive-element-to-noninteractive-role regressions", () => {
+  it("accepts the native-equivalent `cell` role on an interactive `<td>`", () => {
+    const result = runRule(
+      noInteractiveElementToNoninteractiveRole,
+      `<td role="cell" tabIndex={0} onMouseDown={handleMouseDown} onKeyDown={handleKeyDown}>Value</td>`,
+    );
+
+    expect(result.diagnostics).toEqual([]);
+  });
+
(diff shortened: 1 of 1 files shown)


aidenybai:
Closing after runtime-context validation: the proposed allowance would hide a real accessibility defect.

The reported `Cell` component is rendered from `src/components/HighTable/Slice.tsx` under `<table role="grid">`, not a static `<table>`. Under a grid, a native `<td>` has `gridcell` semantics. WAI-ARIA says grid cells have role `gridcell` and the focus target should be a `gridcell`, `rowheader`, or `columnheader`; the APG data-grid guidance likewise says `<td>` implies `gridcell` when its table has `role="grid"`.

The explicit `role="cell"` therefore overrides an interactive grid widget with the static table-cell role. Hightable's `tabIndex`, arrow-key navigation, copy, and mouse/double-click behavior make that distinction material. Its tests querying `role="cell"` demonstrate the current exposed role, but passing tests do not make the ARIA grid contract valid.

A blanket `td: ["cell"]` allowance also demonstrably creates a false negative for `<table role="grid"><tr><td role="cell" tabIndex={0} /></tr></table>`. The existing `<button role="cell">` and mismatched `<td role="listitem">` controls remained reportable, but that is insufficient because the authentic case itself is a true positive.

aidenybai:
No-go: exact runtime composition proves this is a true positive, so I am closing the draft rather than weakening the rule.

PR #1284, opened 2026-07-14
fix(rule): allow validated prop-seeded selection fallback

aidenybai:
## Summary

- reproduce the Semiotic AccessibleNavTree false positive
- preserve a positive control where the prop-seeded state is also rendered without validation

## Current status

Regression-only draft. The valid fixture intentionally fails until the detector change is added.

## Evidence

Benchmark task: `fix-react-rdh-nteract-semiotic-accessiblenavtree`

The unchanged `useState(tree.id)` is local uncontrolled selection. The patched render path retains it only while it belongs to the current tree and otherwise falls back to the current `tree.id`; removing the old reset effect must not turn the unchanged initializer into a new stale-state finding.


aidenybai:
Closing after semantic-proof review. The committed fixture exposes the real benchmark delta, but a safe detector fix must prove the complete controlled-state branch: current-tree provenance, correct memo dependencies, all render-reachable reads, and that `controlledActiveId ?? internalActiveId` cannot select stale internal state. The attempted narrow fallback matcher did not prove the authentic branch, so I am not shipping a heuristic suppression.

PR #1285, opened 2026-07-14
fix(nextjs): allow proven OAuth callback completion redirects

aidenybai:
## Motivation

A browser redirect is the intended completion of many OAuth callback routes. Reporting every such redirect as ordinary client-side navigation pressures users to replace protocol-required behavior, so the exemption must recognize only proven authentication completion flows.

## Example

A callback page may navigate only after the exchange succeeds:

```tsx
const result = await completeOAuthCallback(code)
if (result.ok) {
  window.location.assign(nextDestination)
}
```

Outside a proven auth callback and success path, the existing redirect diagnostic remains.

## Summary

- stop `nextjs-no-client-side-redirect` from reporting browser-owned OAuth callback completion redirects when the route and success proof are statically established
- keep ordinary client redirects diagnosed outside that narrow contract
- add the authentic Bulwark regressions, adversarial near-neighbors, and permanent fuzz coverage

## Precision boundary

The exemption requires an `auth`/`oauth` callback route plus one of three proven completion shapes:

- a successful awaited or immediate promise result gates navigation
- a completion call consumes a direct browser-storage read before navigation
- a storage-backed destination is read, removed, normalized through the authentic browser-navigation helper, and used after successful completion

Unknown calls, transformed promise results, mutations, extra reads, shadowed helpers, ambiguous storage flows, non-callback routes, and ordinary client redirects continue to report.

## Grounded reproduction

- repository: `BulwarkMail/bulwark`
- pinned commit: `c51c3655d5528b10ed15d6449e9f6830597534d5`
- file: `app/(main)/[locale]/auth/callback/page.tsx`
- baseline diagnostics: 3
- candidate diagnostics: 0

The same exact three false positives are removed from the untouched base, benchmark model, and reconstructed oracle. No other diagnostic changes.

## Validation

- focused OAuth regressions: 28/28 pass
- complete plugin suite: 559 files, 14,024 pass, 181 skip
- package typecheck: pass
- repository lint, typecheck, format check, build, JSON smoke, and diff check: pass
- strict generated fuzz: 500 iterations, target fired 10 times, pass
- strict Bench-seeded fuzz: 500 iterations, target fired 10 times, pass
- fuzz corpus suite: 44 pass, 402 skip
- pinned Bulwark base/model/oracle A/B: exactly 3 intended removals, 0 additions, 0 scan errors
- React Bench corpus A/B: 36 files, 0 added, 0 removed, 0 scan errors

The root concurrent `nr test` runner has the same three pre-existing React Native/Expo platform-detection flakes on clean `main` and this branch. This branch adds 13 passing tests and no new failures; the standalone complete plugin suite is green.

Local RDE is not counted as evidence because the installed harness expects report schema v3 while these local artifacts emit v1/v2. The pinned local repository and React Bench comparisons above exercise the AST rule layer directly.

## Changeset

Included for `oxlint-plugin-react-doctor`.


aidenybai:
Closing after detector-design review. The false positive is real and the exact regression is preserved in the branch, but proving browser-owned input -> exact async operation -> resolved success -> navigation requires substantially more bespoke control/dataflow machinery than is justified for one report. The smaller general attempt still exceeded the bounded-change bar, so no implementation commit was made or pushed.

aidenybai:
Closing per the prior detector-design decision: the false positive is preserved, but the current bespoke control/dataflow expansion exceeds the bounded-change bar for this rule. A future fix should use a narrower reusable proof rather than shipping this branch's broad exemption.

PR #1407, opened 2026-07-19
feat(plugin): detect eager Zustand get calls

aidenybai:
## Why

Zustand installs the initial state only after the creator returns. Calling the creator callback's `get` parameter during that eager phase returns `undefined`, even though the public TypeScript type does not expose that possibility.

Bad — the exact helper still executes eagerly:

```ts
const useBearStore = create(async (_set, get) => {
  const readBears = () => get().bears;
  const bears = readBears();
  await ready();
  return { bears };
});
```

Good — defer the read to a store action:

```ts
const useBearStore = create((_set, get) => ({
  bears: 0,
  readBears: () => get().bears,
}));
```

## What the rule reports

`zustand-no-get-during-initialization` reports calls to the proven creator `get` parameter that can execute before the first reachable async suspension.

It covers:

- direct and curried `create`, `createStore`, and `createWithEqualityFn`
- exact official middleware composition
- immutable aliases of `get`
- eager object fields, spreads, computed keys, defaults, and synchronous IIFEs
- exact same-file helper calls, including immutable helper aliases
- the synchronous pre-`await` portion of async creators and async helpers

## Precision boundaries

- A call after a reachable `await` or `for await` suspension stays quiet.
- Deferred actions, methods, nested callbacks, timers, effects, promise callbacks, and generators are not traversed.
- Exact local helper calls are followed only from an eager call site; merely defining or returning a helper does not report.
- Cycles are guarded, generators are skipped, and unresolved/imported helpers fail closed.
- Mutable aliases, shadowed `get` parameters, unknown middleware, and imported custom creators stay quiet.

An async function is not automatically deferred: JavaScript executes its body synchronously until the first suspension. The rule and tests now model that boundary explicitly.

## Version gate

The rule requires `zustand:1`, enabling declared majors 1–5 and disabling missing, unresolved, and future v6 declarations.

## Validation

Current head: `7f1786e0b`.

- Focused rule tests: 21/21.
- Full plugin suite before the final main rebase: 587 files, 16,216 passed / 200 skipped.
- Package typecheck passed after replaying onto current main/foundation.
- Strict targeted fuzz: 500 iterations, no crashes or invariant findings.
- Root foundation gates: 15/15 test tasks, 16/16 typecheck tasks, lint, format check, and JSON smoke.
- Historical exact-SHA Daytona parity for the earlier detector was 100/100 pinned projects, +0/-0, with 3,937 unchanged diagnostics.
- Open-source scans found no eager `get` hits; manually reviewed real-world uses were correctly deferred.

The exact-local-helper false-negative fix postdates the historical parity run, so that result is not presented as current-head parity.

## Exact combined parity

The final five-PR stack was assembled at integration ref `fb20a2ea9b0f1b8722f73197dc7165936d26c6c3` and compared in Daytona with foundation ref `b3e34f271f1e1259b6e2acb006e5b5e5a0a190ae` on the exact same pinned corpus.

- 255 projects from 100 repositories completed on both sides
- zero evaluation failures and zero skipped parity projects
- 27,045 diagnostics unchanged
- zero diagnostics added
- zero diagnostics removed
- parity comparator regression test passed

This exact run includes all four rules enabled together and postdates the final Bugbot fixes for split curried factories, `Object.entries(...).map(...)`, collection-shape proof, and nested notifier ordering.

## Stack

Depends on #1399 for exact Zustand creator and middleware provenance.

---

> [!NOTE]
> **Medium Risk**
> Introduces a new version-gated lint with non-trivial async control-flow analysis and refactors shared suspension logic used by location-invalidation rules; risk is mitigated by broad tests and conservative provenance boundaries.
> 
> **Overview**
> Adds **`zustand-no-get-during-initialization`**, an oxlint rule that flags calls to the creator’s `get` parameter that can run before Zustand has installed initial state (including eager object fields, spreads, sync helpers, and the pre-`await` portion of async creators). Deferred actions, callbacks, and reads after `await` / `for await` are left alone.
> 
> The rule is registered in the plugin, wired through **`createOxlintConfig`** for Zustand majors **1–5** (off when Zustand is missing, unparseable, or v6+), and covered by focused rule tests, liveness/fuzz corpus entries, and verdict-robustness auditing.
> 
> **`canExecuteBeforeAsyncSuspension`** is extracted into a shared util (with `for await` post-suspension regions), and location-invalidation analysis is updated to use it instead of inlined logic; small helpers **`isDescendantWithoutFunctionBoundary`** and **`findImmediatelyInvokedCallExpression`** are shared with **`isImmediatelyInvokedFunction`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8300e9f4bcf8f7e7d782821fbce66979550981a9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aidenybai:
Addressed both findings in `ad4fa47de`.

- async creators now analyze the synchronous prelude through the first suspension boundary, including await operands, while post-await and `for await` bodies stay quiet
- generator creators remain skipped because invoking them does not execute their body
- transparent callee wrappers are unwrapped, covering parenthesized, non-null, and TypeScript-asserted `get` calls
- added three regression groups; focused tests are 20/20, package typecheck passes, plugin build passes, and 500 strict fuzz iterations pass

PR #1540, opened 2026-07-31
fix: [server-auth-actions] skip credential-establishing endpoints

### LINKED ISSUE #1538: [server-auth-actions] Fires on credential-establishing endpoints that the rule's own docs name as its key false positive
**Version:** react-doctor 0.9.2

### Summary

`server-auth-actions` fires on sign-up / OTP-verify / email-verification server actions. The rule's own documentation already identifies these as its key false positive and tells you not to gate them, but the detector still reports them at `error` severity.

From https://react.doctor/docs/rules/react-doctor/server-auth-actions:

> **KEY FALSE POSITIVE** — the action IS the authentication/credential-establishing endpoint, so no prior session can exist: login, signIn, signup/register, OAuth callback, magic-link or OTP verify (verifyAuthCode), password reset/forgot, email verification.

and, in the fix guidance:

> do NOT add a gate to credential-establishing actions (login/signup/OAuth-callback/OTP-verify/password-reset) — those legitimately run for anonymous callers

So the criteria are written down; they just aren't applied. The docs even diagnose why: the credential primitives are "NOT in the helper list, so the rule misfires."

### Reproduction

```ts
"use server";

import { createClient } from "@/lib/supabase/server";
import { checkRateLimit } from "@/lib/rate-limit";

// Reported: "Server action performs unauthenticated privileged server work
// (.update()), so anyone can trigger it directly."
export async function signUpPhoneOnlyAction(input: { phone: string; password: string }) {
  const allowed = await checkRateLimit(`signup:${input.phone}`);
  if (!allowed) return { error: "Too many attempts" };

  const supabase = await createClient();
  const { data, error } = await supabase.auth.signUp({
    phone: input.phone,
    password: input.password,
  });
  if (error) return { error: error.message };

  await supabase.from("profiles").update({ phone: input.phone }).eq("id", data.user!.id);
  return { success: true };
}
```

This is a sign-up endpoint. There is no session to verify, by definition. The `.update()` the message points at operates on the row for the account the call just created.

### Suggested fix

Two options, either would work:

1. Treat `supabase.auth.signUp` / `signInWithPassword` / `signInWithOtp` / `verifyOtp` / `resetPasswordForEmail` (and the equivalents for other auth SDKs) as credential primitives whose presence marks the action as credential-establishing, and skip the rule.
2. Failing that, suppress on the name heuristic the docs already describe: `signUp*`, `signIn*`, `login*`, `register*`, `verify*Otp`, `resetPassword*`.

### Impact

Not a huge deal on its own, but it is `error` severity, and these actions are unfixable by construction. In a codebase with a handful of auth entry points they become permanent errors that no legitimate change clears, which makes the error count less useful as a signal. Suppressing them individually works, but suppressing a rule that the docs say should not have fired feels like the wrong end to fix it from.

Happy to test a patch against a real codebase if that helps.

skoshx:
## Why

`server-auth-actions` reported credential-establishing endpoints such as signup, OTP verification, email verification, OAuth callbacks, and password reset. These endpoints legitimately run for anonymous callers, so requiring a prior session creates an unfixable false positive.

## What changed

- Exempt credential-establishing actions only when a leading credential-intent name is corroborated by an executed known auth SDK call.
- Support direct auth namespaces (`auth.signUp`, `authClient.resetPasswordForEmail`) and provider namespaces such as `supabase.auth.signUp` without accepting unrelated lookalikes such as `analytics.auth.signUp`.
- Normalize camelCase and screaming-snake action names and cover documented aliases including `verifyAuthCode`, `emailVerification`, `passwordReset`, and `handleOAuthCallback`.
- Preserve diagnostics for unrelated privileged actions, non-leading credential words, uncalled helpers, unrelated auth methods, and unknown providers.
- Add a patch changeset for `oxlint-plugin-react-doctor` and `react-doctor`.

## Eval results

| Check             | Result                              |
| ----------------- | ----------------------------------- |
| Projects compared | `2,000`                                            |
| Skipped projects  | `0`                                                |
| Added / removed   | `0 / 1`                                            |
| Target rule delta | `0 / 1`                                            |
| False positives   | `1` confirmed baseline FP removed                  |
| Artifacts         | Daytona `bd4a14ae-b2c9-4979-a164-29e784ad6ed3`; parity SHA-256 `6afd44a9d7ab481994220b074a9ef7121fb9226b3c6cec2577c7d8b101193a1f` |

The sole removal is `antoineross/Hikari@c1f16f54:utils/auth-helpers/server.ts:33`. `signInWithEmail` validates the email, calls `supabase.auth.signInWithOtp`, and writes only a preferred-sign-in-view cookie after the credential call succeeds. This is the intended credential-establishing-endpoint false-positive fix. There were no added diagnostics.

## Test plan

- `nr test src/plugin/rules/server/server-auth-actions.regressions.test.ts` — 84/84 passed
- `nr test` in `packages/oxlint-plugin-react-doctor` — 25,519 passed, 201 skipped
- `FUZZ_RULE=server-auth-actions FUZZ_STRICT=1 FUZZ_ITERATIONS=500 nr fuzz` — passed
- Bounded RDE — 5/5 roots processed, zero `server-auth-actions` hits; JSONL SHA-256 `4f45015866557fbf59ab4e92de0f28bc900f9510df848e90f1b97dd29581b70c`
- `nr typecheck` — 16/16 tasks passed
- `nr format:check` — passed
- `nr test` — 15/15 tasks passed
- `nr lint` — passed (existing warnings only)
- `nr build` — 9/9 tasks passed
- `nr smoke:json-report` — passed (`schemaVersion=3`, full mode)
- Exact one-rule current-main Daytona parity — 2,000/2,000 projects, 0 skipped, 0 added / 1 confirmed-FP removed
- GitHub build, lint, typecheck, CodeQL, React Doctor, package preview, and platform test matrix — passed

Closes #1538

### DIFF
--- a/.changeset/fix-server-auth-actions-credential-establishing.md
+++ b/.changeset/fix-server-auth-actions-credential-establishing.md
@@ -1,0 +1,12 @@
+---
+"oxlint-plugin-react-doctor": patch
+"react-doctor": patch
+---
+
+fix(server-auth-actions): skip credential-establishing actions via SDK detection
+
+The `server-auth-actions` rule now correctly skips server actions that perform credential-establishing operations (signup, signin, OTP verification, password reset) by detecting calls to auth SDK methods like `supabase.auth.signUp()`, `auth.signInWithPassword()`, and `auth.verifyOtp()`. These actions legitimately run for anonymous callers, so requiring authentication would be incorrect.
+
+This resolves the documented false positive where credential-establishing endpoints were incorrectly flagged as unauthenticated privileged operations.
+
+Closes #1538
--- a/packages/oxlint-plugin-react-doctor/src/plugin/rules/server/server-auth-actions.regressions.test.ts
+++ b/packages/oxlint-plugin-react-doctor/src/plugin/rules/server/server-auth-actions.regressions.test.ts
@@ -131,6 +131,68 @@ describe("server/server-auth-actions — regressions", () => {
     expect(result.diagnostics).toEqual([]);
   });
 
+  it("uses the executed auth SDK call when the action name is neutral", () => {
+    const result = runRule(
+      serverAuthActions,
+      `"use server";
+      export async function createAccountAction(input) {
+        const supabase = await createClient();
+        const { data, error } = await supabase.auth.signUp(input);
+        if (error) return { error: error.message };
+        await supabase.from("profiles").update({ ready: true }).eq("id", data.user.id);
+        return { success: true };
+      }`,
+      { filename: "app/actions/auth.ts" },
+    );
+    expect(result.parseErrors).toEqual([]);
+    expect(result.diagnostics).toEqual([]);
+  });
+
+  it("still flags unrelated auth methods", () => {
+    const result = runRule(
+      serverAuthActions,
+      `"use server";
+      export async function deleteAccountAction(input) {
+        const supabase = await createClient();
+        await supabase.auth.signOut();
+        await supabase.from("profiles").delete().eq("id", input.id);
+      }`,
+      { filename: "app/actions/auth.ts" },
+    );
+    expect(result.parseErrors).toEqual([]);
+    expect(result.diagnostics).toHaveLength(1);
+  });
+
+  it("still flags credential calls hidden in uncalled helpers", () => {
+    const result = runRule(
+      serverAuthActions,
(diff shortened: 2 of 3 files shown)


aidenybai:
bugbot run

PR #1541, opened 2026-07-31
fix(effect-needs-cleanup): recognize Supabase channel cleanup

### LINKED ISSUE #1539: [effect-needs-cleanup] supabase.removeChannel(channel) not recognised as cleanup — the recognised alternative is the weaker teardown
**Version:** react-doctor 0.9.2

### Summary

`effect-needs-cleanup` fires on every Supabase Realtime subscription, because the cleanup matcher does not recognise `supabase.removeChannel(channel)`.

The awkward part: the form it *does* recognise, `channel.unsubscribe()`, is the weaker teardown. `unsubscribe()` closes the subscription but leaves the channel registered on the client, so channels accumulate across mounts. `removeChannel()` unsubscribes **and** deregisters, and is what the Supabase docs prescribe. So the rule as written pushes you from the correct teardown toward a leakier one.

### Reproduction

```ts
useEffect(() => {
  const channel = supabase
    .channel(`messages:${roomId}`)
    .on("postgres_changes", { event: "INSERT", schema: "public", table: "messages" }, onInsert)
    .on("postgres_changes", { event: "UPDATE", schema: "public", table: "messages" }, onUpdate)
    .subscribe();

  return () => {
    supabase.removeChannel(channel); // correct teardown, not recognised
  };
}, [roomId, supabase]);
```

Reported as `Effect subscription or timer never cleaned up`, at `error` severity. The subscription *is* cleaned up.

This appears to be the asymmetry the rule docs already describe — the detector descends into nested calls to find the registration, but the cleanup matcher looks for a narrower vocabulary (`removeEventListener`, `clearTimeout`, `unsubscribe`). Suppress criterion 2 covers it in principle:

> a returned cleanup DOES release this resource even if the matcher missed it

but it needs a per-site suppression on every realtime hook in a codebase.

### Suggested fix

Recognise `<client>.removeChannel(<ref>)` in the returned cleanup as releasing a channel created by `<client>.channel(...)` in the same effect. Matching on the method name alone would probably be enough in practice, since `removeChannel` is distinctive.

Two adjacent cases worth folding in, both the documented teardown for their library:

- `supabase.removeAllChannels()`
- `api.dispose()` for Jitsi `JitsiMeetExternalAPI` (registers via `api.addListener(...)`, so the rule fires on the same asymmetry)

### Impact

Any codebase using Supabase Realtime gets one unfixable `error` per subscription hook. We have five, and they are all correct code. Modest in absolute terms, but it is a false positive that scales with how much of the library you use, which is the kind that erodes trust in the error count.

Happy to test a patch.

skoshx:
## Summary

Fixes `effect-needs-cleanup` false positives for Supabase Realtime channels. Closes #1539.

The detector now:

- propagates the assigned terminal `.subscribe()` handle to inner `.on()` calls in a fluent Supabase channel chain;
- accepts `client.removeChannel(channel)` only when both the channel handle and the client that created it match;
- accepts `client.removeAllChannels()` only for channels created by that same client;
- preserves existing `channel.unsubscribe()` and universal lifecycle cleanup such as Jitsi `api.dispose()`.

It deliberately rejects cleanup on a different client, cleanup of a different channel, and unrelated emitter `.on()` calls that merely coexist with a Supabase cleanup call.

## Validation

- Focused rule tests: 492 passed
- Full oxlint plugin: 25,514 passed, 201 skipped
- Full repository suite: 15/15 tasks passed
- Strict target fuzz: 500 iterations passed
- Typecheck: 16/16 tasks passed
- Lint, format check, build (9/9), and JSON schema-v3 smoke: passed
- Post-change symbol/dedup search: no duplicate helper found
- RDE compatibility build: passed; a live RDE run could not start on this host because its required encrypted Axiom environment loader was unavailable

## Exact parity

- Base: `19387631f2c49342c51597445459ce621ff4df21`
- Head: `f4a0c1a1f363408c3eaad3abbe0ca56becd2faa0`
- Scope: `react-doctor/effect-needs-cleanup`
- Projects: 2,000/2,000 compared, 0 failed, 0 skipped
- Diagnostics: `+0/-5`, 2,669 unchanged

All five removals were reconstructed at their pinned source commits. Each is a confirmed baseline false positive with exact owned cleanup: three terminal-handle `unsubscribe()` cases, one same-client `removeChannel(channel)` case, and one guarded ref-retained channel passed to the same client's `removeChannel`. There are zero candidate false positives, removed true positives/current-head false negatives, subjective or out-of-contract rows, reconstruction failures, or persistence gaps.

Artifact SHA-256:

- Base: `b6390d33a5b3f45be8b39f3c41a0b6c15a370178a24e5aa1d9096713a5577af6`
- Candidate: `69e9cebfa0de4b705759f93edae32bdec055d84b6f8b914ac9b6109b05f91b31`
- Parity: `de0accc03c6e7cfb03ec3cb040c3e453f20477128d9d0df7713a9cdfd9f5b6c6`
- Provenance: `17e80ffe97438919472d195afaa957dca5288e05fff0b12ffdd87e969a3e9cd7`
- Final adjudication: `c6690e2e4428f5f6fd8de0825b9cfc3b1f978ffa11ec5e8a368cc22c66240a3b`

Exact Daytona cleanup was independently verified: zero evaluation-labeled sandboxes remain and the exact snapshot is absent. Current-head CI and Cursor Bugbot are green, with zero unresolved review threads.

### DIFF
--- a/.changeset/supabase-removechannel.md
+++ b/.changeset/supabase-removechannel.md
@@ -1,0 +1,15 @@
+---
+"oxlint-plugin-react-doctor": patch
+---
+
+fix(effect-needs-cleanup): recognize Supabase removeChannel/removeAllChannels and Jitsi dispose cleanup
+
+The rule now correctly recognizes:
+- `supabase.removeChannel(channel)` as valid cleanup for Supabase Realtime subscriptions
+- `supabase.removeAllChannels()` as global cleanup for all Supabase channels
+- `api.dispose()` as valid cleanup for Jitsi ExternalAPI (already worked via UNIVERSAL_RELEASE_VERB_NAMES)
+- `channel.unsubscribe()` for Supabase channels created with `.on()` chains
+
+This eliminates false positives when using the documented Supabase cleanup methods, which are stronger than the previously-recognized `channel.unsubscribe()` pattern (removeChannel both unsubscribes and deregisters the channel).
+
+Fixes #1539
--- a/packages/oxlint-plugin-react-doctor/src/plugin/constants/react.ts
+++ b/packages/oxlint-plugin-react-doctor/src/plugin/constants/react.ts
@@ -198,8 +198,6 @@ export const GLOBAL_RELEASE_METHOD_NAMES = new Set([
   "disconnect",
   "unobserve",
   "close",
-  "removeChannel",
-  "removeAllChannels",
 ]);
 
 export const BOUND_RESOURCE_RELEASE_METHOD_NAMES = new Set([
--- a/packages/oxlint-plugin-react-doctor/src/plugin/constants/react.ts
+++ b/packages/oxlint-plugin-react-doctor/src/plugin/constants/react.ts
@@ -198,6 +198,9 @@ export const GLOBAL_RELEASE_METHOD_NAMES = new Set([
   "disconnect",
   "unobserve",
   "close",
+  // Supabase Realtime channel removal (parent-level cleanup for subscriptions).
+  "removeChannel",
+  "removeAllChannels",
 ]);
 
 export const BOUND_RESOURCE_RELEASE_METHOD_NAMES = new Set([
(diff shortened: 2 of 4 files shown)


aidenybai:
bugbot run

PR #1552, opened 2026-08-02
FPFNFIX82 Recognize media failure state lifecycles

aidenybai:
## Why\n\nProp-keyed media failure latches are external-resource lifecycle state, not redundant render-derived state. React Doctor rejected valid retry behavior when video source or MIME identity changed.\n\n## What changed\n\n- prove that the same setter is written by media load, error, or nested rejected-play handlers\n- exempt only that proven resource lifecycle from no-adjust-state-on-prop-change\n- exempt no-reset-all-state-on-prop-change only when the component has one proven resource-owned state cell\n- preserve ordinary draft, selection, playability, and multi-state reset diagnostics\n- add focused regressions and fuzz fixtures\n\n## Eval results\n\nIn the cached Jumper replay, 77 of 187 cached-new no-adjust trial findings and 44 of 49 no-reset findings were removed. Current replay retained 118 no-adjust and 5 no-reset diagnostics for patterns outside the proof.\n\nPull request parity is blocked because DAYTONA_API_KEY is unavailable in the publishing environment.\n\n## Test plan\n\n- nr -C packages/oxlint-plugin-react-doctor test\n- nr -C packages/fuzz test\n- strict 500-iteration fuzz runs for both changed rules\n- nr test\n- nr typecheck\n- nr lint\n- nr format:check


aidenybai:
Superseded by #1546 and #1547, which implement the same media failure-latch fixes for both affected rules with stricter intrinsic-resource provenance, broader counterexample coverage, fuzz coverage, and completed 5,766-project parity.

PR #1553, opened 2026-08-02
FPFNFIX82 Preserve effect-chain diagnostics with cleanup

aidenybai:
## Why\n\nno-effect-chain treated any cleanup return as proof that an entire effect was external synchronization. Unrelated object URL cleanup and ref bookkeeping could therefore hide a real state-to-state effect chain.\n\n## What changed\n\n- keep analyzing synchronous state work when an effect also returns direct cleanup\n- follow state through immutable render expressions and memo dependencies\n- prevent ref-only bookkeeping from laundering opaque callback chains\n- preserve genuine subscriptions, DOM synchronization, storage writes, and helper-owned cleanup\n- add focused regressions and a true-positive fuzz fixture\n\n## Eval results\n\nThe cached Inrupt confirmed cluster changed from 0 of 153 detected to 153 of 153. The broader candidate replay detects 169 of 175; the remaining six have no local state chain.\n\nPull request parity is blocked because DAYTONA_API_KEY is unavailable in the publishing environment.\n\n## Test plan\n\n- nr -C packages/oxlint-plugin-react-doctor test\n- nr -C packages/fuzz test\n- FUZZ_RULE=no-effect-chain FUZZ_ITERATIONS=500 FUZZ_STRICT=1 nr -C packages/fuzz fuzz\n- nr test\n- nr typecheck\n- nr lint\n- nr format:check


aidenybai:
Superseded by #1545, which implements the same mixed effect-chain fix with CFG reachability, preserves additional subscription/DOM/canvas/channel lifecycles, adds broader regression and fuzz coverage, and includes completed paired parity.

PR #1555, opened 2026-08-02
FPFNFIX82: recognize self-reconnecting observer cleanup

aidenybai:
## Why

ReactBench RDH exposed an `effect-needs-cleanup` false positive in Cloudscape's visual-context observer (`write-react-cloudscape-design-co__BkahTcP`).

The effect owns one `MutationObserver` variable. Every reconnect releases the previous observer before replacing it, and the returned cleanup releases the latest observer on unmount:

```tsx
let observer: MutationObserver | undefined

const reconnect = () => {
  observer?.disconnect()
  observer = new MutationObserver(reconnect)
  observer.observe(element, { attributes: true, childList: true })
}

reconnect()
return () => observer?.disconnect()
```

Before this change, the detector anchored the nested `observe()` call inside `reconnect` and could not prove that the outer direct invocation was followed by cleanup. Rewriting this working lifecycle solely to satisfy the diagnostic adds churn without improving resource ownership.

The platform contract says `disconnect()` stops the observer and permits later reuse. React requires Effect cleanup to undo setup on dependency change or unmount. This pattern satisfies both contracts.

- MDN: https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/disconnect
- React: https://react.dev/reference/react/useEffect#reference

## What changed

- Recognize only a self-reconnecting helper whose observer binding is local and nullable.
- Require exactly one observer construction and a guaranteed matching release before every replacement.
- Require exactly one direct helper invocation and the normal returned Effect cleanup for the latest observer.
- Reject async/generator helpers, conditional replacement cleanup, extra retainers, unknown observer uses, and replacement without prior release.
- Add the exact ReactBench shape, adversarial negative cases, and a fuzz regression.

The existing true-positive behavior is preserved: replacement without `observer?.disconnect()` before construction still reports.

## Eval results

A bounded local RDE pass processed 25 project roots. The target rule produced only its existing real-world diagnostic population; this branch adds no new diagnostic surface. A wider pass reached 99 roots but was interrupted on a stalled monorepo root and is not counted as parity evidence. Daytona PR parity will be attached separately.

## Test plan

- `nr test --filter oxlint-plugin-react-doctor`
- `nr typecheck --filter oxlint-plugin-react-doctor`
- `nr build`
- `FUZZ_RULE=effect-needs-cleanup FUZZ_STRICT=1 FUZZ_ITERATIONS=500 nr fuzz`
- `nr test && nr lint && nr typecheck && nr format:check && nr smoke:json-report`

---

> [!NOTE]
> **Medium Risk**
> Changes static analysis heuristics in a widely used effect rule; incorrect recognition could hide real leaks or reintroduce false positives, though scope is narrow and heavily regression-tested.
> 
> **Overview**
> Fixes a **false positive** in `effect-needs-cleanup` when an effect uses a **self-reconnecting observer helper** (disconnect → `new MutationObserver` → observe) with unmount cleanup on the latest instance—matching Cloudscape’s visual-context pattern.
> 
> The rule now treats that helper like a single setup path: **`findReconnectHelperInvocation`** proves a local nullable `let` observer, one reassignment to a global observer constructor, **mandatory `disconnect` before every replacement**, exactly one direct helper call from the effect, and the usual returned cleanup. **`resolveCleanupPathAnchor`** anchors cleanup coverage on that invocation instead of the nested `observe()` call.
> 
> **Still flagged:** replacement without prior release, conditional disconnect, extra retainers of the helper, and other adversarial shapes. Adds regression tests, a fuzz corpus case, and a patch changeset for both lint plugins.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5144487f96706e5e52218b03b74b51a37ca9f6ec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/recognize-reconnecting-observer-cleanup.md
+++ b/.changeset/recognize-reconnecting-observer-cleanup.md
@@ -1,0 +1,6 @@
+---
+"oxlint-plugin-react-doctor": patch
+"eslint-plugin-react-doctor": patch
+---
+
+Recognize observer reconnect helpers that release each replacement and the latest observer on cleanup.
--- a/packages/fuzz/corpus/regressions/effect-needs-cleanup--reconnecting-observer.tsx
+++ b/packages/fuzz/corpus/regressions/effect-needs-cleanup--reconnecting-observer.tsx
@@ -1,0 +1,16 @@
+// effect-needs-cleanup | wrapper-transparency | ReactBench Cloudscape visual context
+import { useLayoutEffect } from "react";
+
+export const VisualContextWatcher = ({ element }: { element: HTMLElement }): null => {
+  useLayoutEffect(() => {
+    let observer: MutationObserver | undefined;
+    const refresh = () => {
+      observer?.disconnect();
+      observer = new MutationObserver(refresh);
+      observer.observe(element, { attributes: true, childList: true });
+    };
+    refresh();
+    return () => observer?.disconnect();
+  }, [element]);
+  return null;
+};
--- a/packages/oxlint-plugin-react-doctor/src/plugin/rules/state-and-effects/effect-needs-cleanup.regressions.test.ts
+++ b/packages/oxlint-plugin-react-doctor/src/plugin/rules/state-and-effects/effect-needs-cleanup.regressions.test.ts
@@ -3001,6 +3001,96 @@ export const useResizableColumns = () => {
 });
 
 describe("effect-needs-cleanup adversarial edge cases (observers / connections / retained functions)", () => {
+  it("accepts an observer replaced by a self-reconnecting helper with unmount cleanup", () => {
+    const result = runRule(
+      effectNeedsCleanup,
+      `import { useLayoutEffect } from "react";
+export const ContextWatcher = ({ element }) => {
+  useLayoutEffect(() => {
+    let observer;
+    const refresh = () => {
+      observer?.disconnect();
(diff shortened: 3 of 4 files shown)


aidenybai:
## Daytona parity

:warning: **Blocked, not passed.**

The batch-efficient scoped matrix used exact base `b02bc694f134fc856ad1e17304a93e0aba3e31a6`, head `5144487f96706e5e52218b03b74b51a37ca9f6ec`, and all 5,766 pinned project roots. After four retries (concurrency 50 → 10 → 2), the candidate and shared base each retained the same 14 failed roots; all are `1771-Technologies/lytenyte` documentation examples and all ended with `Evaluation time budget exhausted`.

- Candidate records: 5,766 total, 5,752 complete, 14 failed
- Baseline: unavailable because the shared base lane failed the same tail
- Comparator: deliberately not run against incomplete inputs
- Candidate artifact: `tmp/parity-batch-1555-1557-scoped-b02bc694f/pr-1555/artifacts/`
- Evaluation ID: `8c62b86b-fa3a-470f-89c2-531281c80210`

This is a corpus/runtime blocker shared by base and treatment, not evidence of a detector regression. CI, focused regressions, strict fuzzing, and the bounded local RDE pass remain green.

PR #1596, opened 2026-08-07
fix(rule): harden effect cleanup regressions

### LINKED ISSUE #1593: False Positive: react-doctor/no-loading-flag-reset-outside-finally
**Rule:** `react-doctor/no-loading-flag-reset-outside-finally`

### Summary

The rule flags a `setLoading(false)` placed after a `try/catch` block as "reset only on the success path." But when the `catch` clause handles the error **without re-throwing**, execution falls through to the trailing reset on *both* the success and the caught-rejection paths. The flag is never stuck truthy, so there is no bug — this is functionally equivalent to a `finally` reset.

The rule's detector appears to only check whether the reset is lexically inside `finally`, without tracing that the trailing statement is reachable from a non-rethrowing `catch`.

### Actual Code

```tsx
  const handleUpload = async () => {
    if (!selectedFile || isUploading) return;
    setIsUploading(true);

    try {
      const mimeType = selectedFile.mimeType ?? "application/octet-stream";
      const formData = new RawNameFormData();
      const filePart = {
        uri: selectedFile.uri,
        name: selectedFile.name,
        type: mimeType,
      };
      // oxlint-disable-next-line typescript/no-unsafe-type-assertion
      formData.append("file1", filePart as unknown as Blob);

      const res = await authedFetch(`${API_BASE}/lectures/${lectureId}/tasks/${taskId}/submit`, {
        method: "POST",
        body: formData,
      });

      if (res.ok) {
        onSuccess();
        setSelectedFile(null);
      } else {
        toast.show({ variant: "danger", label: t("lectures.grades.uploadError"), duration: 3500 });
      }
    } catch {
      toast.show({ variant: "danger", label: t("lectures.grades.uploadError"), duration: 3500 });
    }
    setIsUploading(false);
  };
```

**Actual:** `no-loading-flag-reset-outside-finally` fires on the trailing `setIsUploading(false)`.
**Expected:** no diagnostic — the reset is guaranteed to run on rejection because the `catch` does not re-throw.

### Why this matters — conflict with `react-hooks-js/todo`

The suggested fix (move the reset into `finally`) is not viable in a React Compiler codebase. `react-hooks-js/todo` reports:

> Todo: (BuildHIR::lowerStatement) Handle TryStatement without a catch clause

React Compiler cannot currently lower `try`/`finally` (or `try` without `catch`). So the two rules are in direct conflict:

- `no-loading-flag-reset-outside-finally` → wants `finally`
- `react-hooks-js/todo` → forbids `finally`

The `catch`-without-rethrow + trailing-reset pattern is the *only* shape that satisfies both, yet this rule flags it.

### Environment

- "react-doctor": "0.9.5"
- React Compiler / `react-hooks-js` enabled: yes
- React 19, React Native 0.85 (Expo 56) (React Compiler enabled)

### LINKED ISSUE #1594: effect-needs-cleanup false-positives when the timer is cleared via a helper called from the cleanup return
**Rule:** `react-doctor/effect-needs-cleanup`
**Version:** 0.9.5

### Summary

The rule flags a `setTimeout` inside `useEffect` as "never cleaned up" even when its handle **is** cleared in the effect's cleanup `return` — just indirectly, through a local helper function rather than an inline `clearTimeout`. The detector's cleanup matcher misses the helper indirection.

### Reproduction

```tsx
useEffect(() => {
  const timerRef = { current: null as ReturnType<typeof setTimeout> | null };

  const clearOfflineTimer = () => {
    if (timerRef.current) {
      clearTimeout(timerRef.current);
      timerRef.current = null;
    }
  };

  const applyState = (online: boolean) => {
    if (!online && !timerRef.current) {
      timerRef.current = setTimeout(() => { /* ... */ }, GRACE_MS); // ⚠️ flagged
    }
  };
  // ...

  return () => {
    clearOfflineTimer(); // <-- timer IS released here
    // ...other teardown
  };
}, []);
```

**Actual:** `effect-needs-cleanup` fires on the `setTimeout`.
**Expected:** no diagnostic — the handle is stored in a ref and cleared on teardown via `clearOfflineTimer()`.

aidenybai:
## Why

`no-loading-flag-reset-outside-finally` treated every opaque call in a swallowing `catch` as a possible synchronous throw, so handled upload failures could incorrectly make a trailing loading reset look unreachable. The `effect-needs-cleanup` local-helper shape from #1594 was already quiet on current source, but its ownership boundary was not protected by deterministic semantic fuzzing.

After this change, ordinary error-reporting calls can fall through to the reset, while explicit exits and exact local helpers proven to throw on every CFG path remain reportable. Cleanup helper indirection is covered by both pass and fail seeds instead of relying on random mutation to rediscover it.

## What changed

- Narrow catch bypass detection to CFG-proven always-throwing local helpers and cache repeated helper coverage.
- Replay labeled fuzz corpus verdicts deterministically and report mismatches as blocking findings.
- Add an AST-equivalent cleanup-helper mutation for `effect-needs-cleanup`.
- Add focused FP/FN regressions for #1593 and #1594 plus a patch changeset.

## Eval results

| Check | Result |
| --- | --- |
| Project roots compared | 100 |
| Skipped projects | Not reported by the local runner |
| Added / removed | 0 / 0 |
| Target rule delta | 0 / 0 |
| False positives | 0 in inspected target-rule hits |
| Artifacts | Local RDE comparison against `react-doctor@0.9.5` |

## Test plan

- `nr test` — 15/15 tasks passed
- `nr lint` — passed
- `nr typecheck` — 16/16 tasks passed
- `nr format:check` — passed
- `nr smoke:json-report` — schema v3 smoke passed
- `FUZZ_RULE=effect-needs-cleanup FUZZ_STRICT=1 FUZZ_ITERATIONS=500 nr fuzz` — passed
- `FUZZ_RULE=no-loading-flag-reset-outside-finally FUZZ_STRICT=1 FUZZ_ITERATIONS=500 nr fuzz` — passed
- `nlx react-doctor@latest --verbose --scope changed` — 100/100, no diagnostics

Closes #1593.
Closes #1594.

---

> [!NOTE]
> **Medium Risk**
> Changes control-flow analysis in a state/effects lint rule and tightens fuzz oracles; behavior shifts are intentional but could affect edge-case diagnostics beyond the fixed issues.
> 
> **Overview**
> **`no-loading-flag-reset-outside-finally`** no longer treats every opaque call in a swallowing `catch` as blocking a trailing loading reset. On paths that continue to the reset, only **CFG-proven always-throwing** exact local helpers still count as abrupt completion; ordinary reporting calls (e.g. `toast.show`) allow the reset. Tests and corpus cover #1593.
> 
> **Fuzz harness** parses `// rule:` / `// verdict:` from corpus files, **replays declared pass/fail verdicts** at fuzz start (new **`verdict-mismatch`** finding), and documents that labeled seeds run in smoke and targeted fuzz. For **`effect-needs-cleanup`**, an AST variant hoists effect cleanup calls into local helpers (with guards so cleanup-local bindings are not hoisted); pass/fail seeds lock #1594 behavior.
> 
> **`effect-needs-cleanup`** gains focused unit tests for effect-local cleanup helpers (subscription vs deferred promise allocation).
> 
> Patch changeset for `react-doctor` and `oxlint-plugin-react-doctor`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a231dc0537197844ca2c4762392e15e405689f8b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/quiet-swallowed-catches.md
+++ b/.changeset/quiet-swallowed-catches.md
@@ -1,0 +1,6 @@
+---
+"oxlint-plugin-react-doctor": patch
+"react-doctor": patch
+---
+
+Accept trailing loading resets after non-rethrowing catch handlers, and preserve effect-local cleanup helpers with deterministic false-positive and false-negative fuzz replay.
--- a/packages/fuzz/README.md
+++ b/packages/fuzz/README.md
@@ -45,10 +45,10 @@ that never fires is only having its early bails fuzzed.
 
 Each compatible canonical liveness fixture runs first, so the harness reaches
 a known reporting path before exploring generated and corpus-derived programs.
-Corpus seeds with a `// verdict: pass` or `// verdict: fail` header are also
-run deterministically by the smoke suite against the rule named in their
-`// rule:` header. Use `pass` for false-positive regressions and `fail` for
-confirmed true positives.
+Corpus seeds with a `// verdict: pass` or `// verdict: fail` header run
+deterministically in both the smoke suite and the targeted fuzz run against the
+rule named in their `// rule:` header. Use `pass` for false-positive regressions
+and `fail` for confirmed true positives.
 
 Every case is reproducible from its seed; reproducers for findings are written
 to `tmp/fuzz-findings/`.
--- a/packages/fuzz/corpus/README.md
+++ b/packages/fuzz/corpus/README.md
@@ -12,11 +12,10 @@ Two seed families, split by expected rule verdict:
   genuine bug a rule must flag, kept as a mutation seed so its shape keeps
   applying pressure.
 
-The harness enforces **no firing expectations** for either family — its
-oracles are crash, slowness, verdict-preserving invariance, and verdict
-drops, and `firedProgramCount` is a coverage stat only. Whether a seed must
-or must not fire is pinned by the owning rule's unit test file, never by
-fuzzing.
+Seeds with `// verdict: pass` or `// verdict: fail` are replayed
+deterministically by both the smoke suite and targeted fuzzing. Seeds without a
+verdict remain mutation-only inputs. `firedProgramCount` is still a coverage
+stat rather than a correctness oracle for generated programs.
 
 **The evolving loop (see the `fuzz` skill):** whenever a new false positive
 is confirmed — from a user report, an RDE eval, a react-bench run, review,
--- a/packages/fuzz/corpus/regressions/effect-needs-cleanup--issue-1594-local-cleanup-helper.tsx
+++ b/packages/fuzz/corpus/regressions/effect-needs-cleanup--issue-1594-local-cleanup-helper.tsx
@@ -1,0 +1,30 @@
+// verdict: pass
+// rule: effect-needs-cleanup
+// weakness: cleanup-provenance
+// source: issue #1594
+
+import { useEffect } from "react";
+
+export const ConnectionStatus = ({ online }) => {
+  useEffect(() => {
+    const timerRef = { current: null as ReturnType<typeof setTimeout> | null };
+    const clearOfflineTimer = () => {
+      if (timerRef.current) {
+        clearTimeout(timerRef.current);
+        timerRef.current = null;
+      }
+    };
+    const applyState = (nextOnline: boolean) => {
+      if (!nextOnline && !timerRef.current) {
+        timerRef.current = setTimeout(() => updateStatus(), 1_000);
+      }
+    };
+
+    applyState(online);
+    return () => {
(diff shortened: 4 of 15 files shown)


aidenybai:
Daytona pushed-head parity could not run from this environment because `DAYTONA_API_KEY` is unavailable. The PR does not claim Daytona parity. The local RDE evidence in the description compares 100 project roots against `react-doctor@0.9.5` with zero target-rule diagnostic delta.

PR #1654, opened 2026-08-16
feat(rules): add UI-library composition and correctness rules behind capability gates

aidenybai:
## Why

UI libraries carry composition and stability contracts that plain JSX linting can't see: the part names and option semantics only mean something in a project that actually uses that library. This PR adds seven project capabilities — `shadcn` (from `components.json`), `radix-ui` (unified `radix-ui` package or any `@radix-ui/react-*` primitive, icons excluded), `base-ui` (`@base-ui-components/react` or its 1.0 rename `@base-ui/react`), `react-aria` (`react-aria-components`), `tanstack-table`, `tanstack-virtual`, and `tanstack-form` — plus thirteen rules, each resolving components and hooks through their imports (never by name alone) and enabled only under its library's gate.

### Unnamed dialogs (Accessibility)

A dialog surface without a title part and without `aria-label` has no accessible name — assistive technology announces an unnamed dialog, and Radix logs a runtime accessibility error.

- **`shadcn-dialog-content-requires-title`** — `DialogContent`, `SheetContent`, `AlertDialogContent`, `DrawerContent`
- **`radix-dialog-content-requires-title`** — `Dialog.Content`, `AlertDialog.Content` (per-primitive packages and the unified `radix-ui` package)
- **`base-ui-dialog-popup-requires-title`** — `Dialog.Popup`, `AlertDialog.Popup` (both package names, subpath and root imports)
- **`react-aria-dialog-requires-heading`** — React Aria `Dialog` without a `Heading` (headings inside the common `{({ close }) => …}` render-prop child are found before reporting)

```tsx
// Before: announced as an unnamed dialog
<Dialog.Content>
  <p>Are you sure?</p>
</Dialog.Content>

// After
<Dialog.Content>
  <Dialog.Title>Delete file</Dialog.Title>
</Dialog.Content>
```

The detectors stay quiet whenever the subtree is not statically provable: spread props, name-providing attributes, `{children}` or any non-literal expression child, `render` props, unresolved custom components, and (for shadcn) same-module `*Header` parts with content, since some shadcn-compatible distributions render the header text as the title. Titles supplied through `render={<DialogTitle />}` composition and inside wrappers or map callbacks are found before reporting.

### Unlabeled fields and buttons (Accessibility)

- **`shadcn-icon-button-requires-label`** — icon-sized `Button`s whose statically-known content is only icons. sr-only text and `<svg><title>` count as names; `aria-hidden` content does not; buttons composed through `render` props (label lives in the wrapper's children) and react-aria `slot`s (the context names them, e.g. TagGroup's remove slot) are skipped.
- **`shadcn-form-item-requires-label`** — `FormItem` wrapping a `FormControl` without a `FormLabel` or aria naming. The canonical react-hook-form `{...field}` spread stays checkable; any other spread quiets the rule.
- **`base-ui-field-requires-label`** — `Field.Root` wrapping a `Field.Control` without a `Field.Label`, intrinsic label, or aria naming — including names supplied through Base UI's `render` prop.

### Broken compositions and option contracts (Bugs)

- **`tanstack-table-no-unstable-data-or-columns`** — the table docs treat stable `data`/`columns` references as a requirement: a fresh array identity per render rebuilds every row/column model each render, and auto-reset features then loop state updates forever. Reports only provably-fresh identities (inline array literals, render-scoped `const` arrays, fresh `?? []` / ternary fallbacks, inline `.filter()`/`.map()` transforms). Covers `useReactTable` (v8) and `useTable` (v9).
- **`tanstack-virtual-measure-element-requires-data-index`** — virtual-core's `indexFromElement` reads `data-index`; a `ref={virtualizer.measureElement}` element without it gets its measurement dropped with a console warning. Covers member, destructured, and callback ref forms; quiet under a custom `indexAttribute`.
- **`tanstack-form-on-submit-requires-prevent-default`** — TanStack Form's `handleSubmit` never cancels the native submit, so `onSubmit={form.handleSubmit}` (or an inline wrapper that skips `event.preventDefault()`) performs a full-page form submission mid-submit. The docs mandate the preventDefault wrapper.
- **`shadcn-input-group-no-raw-controls`** — raw `Input`/`Textarea`/`Button` directly inside `InputGroup` double-frames the control while the group's shared focus/error styling never applies.
- **`shadcn-command-item-state-variant-requires-value`** — cmdk renders `data-selected`/`data-disabled` on every item as `"true"` **or** `"false"`, so a presence-only `data-[selected]:` variant styles every item.
- **Tabs triggers outside the list** — `shadcn-tabs-trigger-requires-list` (existing rule, now capability-gated and hardened), **`radix-tabs-trigger-requires-list`**, and **`base-ui-tabs-tab-requires-list`**. All three report only a trigger *provably inside the root without crossing the list part*.

## Fuzzing and real-world false-positive hunt

- **Strict fuzzing**: every new rule ran through the adversarial harness with `FUZZ_STRICT=1 FUZZ_REQUIRE_FIRE=1 FUZZ_ITERATIONS=600` — all fired (reporting paths exercised, not just early bails) with zero crash, slow, invariant-violation, or verdict-drop findings; modified rules were re-fuzzed after the FP fixes.
- **Real-code hunt**: all rules ran over a 5,067-file corpus of seven repos that use these libraries heavily (shadcn-ui/ui, TanStack table/virtual/form, radix-ui/primitives, Intent UI, and a production shadcn app). Four false-positive classes were found, fixed, and pinned with regression tests plus fuzz-corpus fixtures:
  1. childless `Button` composed via a `render` prop, with the sr-only label in the wrapper's children;
  2. titles supplied through `render={<DialogTitle />}` composition (shadcn v4 questionnaire dialogs);
  3. shadcn-compatible distributions whose `*Header` part renders the title text itself (Intent UI drawers);
  4. react-aria slotted buttons (`slot="remove"`) named by their surrounding context.
- After the fixes, every remaining corpus hit was manually triaged as a true positive (untitled demo drawers and genuinely unnamed icon buttons in example code); the other eleven rules produced zero diagnostics on valid corpus code.

## Research verified against primary sources

Contracts were checked against the shadcn registry, Radix/Base UI/React Aria docs and npm exports, cmdk's published source, TanStack Table's data guide, TanStack Form's quick start, and `@tanstack/virtual-core`'s implementation. Candidates **rejected** during verification: `manualSorting` + client row model (v9 docs explicitly endorse the combination), controlled `state.X` without `onXChange` (externally-driven state is a legitimate pattern), TanStack required-option checks (TypeScript enforces them), and React Aria `onClick`-vs-`onPress` (recent versions alias `onClick`).

## What changed

- Seven capabilities surfaced through `ProjectInfo` and `buildCapabilities` (components.json detection plus dependency any-of predicates with a shared `hasAnyDependency` helper).
- Shared detection utilities: `resolveShadcnUiComponentName`, `resolveNamespacedPartName`, `visitStaticJsxChildren`, `scanJsxSubtreeForPart` (now also searching `render` props), `findRequiredAncestorPlacement`, `getValueImportDeclaration`, `getTrailingJsxNameSegment`. All thirteen rules share one detection core.
- Registry regenerated (906 rules) with liveness fixtures for every new rule; ESLint presets pick everything up through the generated registry; two fuzz regression fixtures.
- Patch changesets for `oxlint-plugin-react-doctor`, `eslint-plugin-react-doctor`, and `react-doctor`.

## Non-goals (v1)

- Descendant-scoped `group-data-[selected]/…` variants on command-item children (needs group-name resolution).
- Titles/labels delegated to unresolved custom components are trusted, not followed cross-file.
- Icon-only buttons at the default size; `FormMessage`/`Field.Error` presence.
- `@radix-ui/themes`, Radix `Form`, TanStack Router-only projects (the tanstack-start family is framework-gated), and TanStack Query `setQueryData` cache mutation — separate contracts worth their own evidence pass.

## Test plan

- `pnpm test` — all 12 workspace tasks pass; `pnpm typecheck`, `pnpm lint`, `pnpm format`, `pnpm smoke:json-report` — all pass.
- Focused: sixteen rule test files plus liveness, and `packages/core` capability/detection tests.
- Fuzz + corpus-hunt results as above.

<div><a href="https://cursor.com/agents/bc-330a2bd5-d81a-457c-afbc-4738effa42da?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-330a2bd5-d81a-457c-afbc-4738effa42da&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


aidenybai:
bugbot run

PR #1662, opened 2026-08-20
fix(eslint-plugin): honor disabledWhen capabilities in ESLint plugin

### LINKED ISSUE #1661: eslint-plugin-react-doctor ignores `disabledWhen: ['react-compiler']`
The CLI drops the rules tagged `disabledWhen: ['react-compiler']` when it detects the
compiler. The ESLint plugin doesn't, so they all still fire in the editor.

Same repo (Next.js 16, React 19, `reactCompiler: true`), all packages 0.9.12:

    CLI  ....... 0 × jsx-no-new-{function,array,object,jsx}-as-prop
    ESLint ..... 1,998 of them (recommended + next + tanstack-query)

Cause — `eslint-plugin-react-doctor/dist/index.js`:

    const wrapAsEslintRule = (ruleName, ruleImpl) => ({
      meta: { type: ..., docs: {...}, schema: [] },
      create: ruleImpl.create          // disabledWhen never read
    });

`disabledWhen` appears 0 times in that file; there's no `settings` hook and no
compiler-aware preset. The CLI resolves the gate from `ProjectInfo.hasReactCompiler`.

This also makes `recommended` self-contradictory: the `useCallback` that silences
`jsx-no-new-function-as-prop` is then reported by `react-compiler-no-manual-memoization`.
One file of ours: 15 warnings before that fix, 15 after. Named handlers don't help — the
detector resolves render-local identifiers.

Ask: have `wrapAsEslintRule` honour `disabledWhen`, opted into explicitly since ESLint
can't read `next.config.ts`:

    settings: { 'react-doctor': { reactCompiler: true } }

A `configs['recommended-compiler']` preset would work too.

skoshx:
Fixes #1661

## Root Cause

The ESLint plugin wrapper was not checking rule.disabledWhen metadata. Rules like jsx-no-new-function-as-prop that have disabledWhen: ['react-compiler'] kept firing in ESLint even when React Compiler was configured.

The CLI correctly disabled these rules via the oxlint plugin's shouldCreateRuleVisitors check, but the ESLint wrapper just unconditionally called ruleImpl.create.

## Scope

The fix is narrowly scoped to the ESLint plugin wrapper only:

- Import capability checking helpers from oxlint-plugin-react-doctor
- Check context.settings['react-doctor'].capabilities against rule.disabledWhen
- Return empty visitors when any disabledWhen capability is present
- This matches the oxlint plugin's behavior

Affects 18 rules with disabledWhen: ['react-compiler']:
- jsx-no-new-{function,array,object,jsx}-as-prop
- jsx-no-constructed-context-values
- context-provider-value-from-unmemoized-local-literal
- no-inline-prop-on-memo-component
- prefer-module-scope-{static-value,pure-function}
- prefer-stable-empty-fallback
- rendering-hoist-jsx
- rerender-{memo-with-default-value,dependencies}
- no-effect-with-fresh-deps
- rn-no-inline-{object-in-list-item,flatlist-renderitem}
- rn-list-{data-mapped,callback-per-row}

## Usage

Users enable this by adding to their ESLint config:

```js
export default {
  settings: {
    'react-doctor': {
      capabilities: ['react-compiler']
    }
  }
}
```

## Testing

- Added comprehensive test coverage for all 18 affected rules
- Verified rules return empty visitors when capability is present
- Verified rules return populated visitors when capability is absent
- Verified behavior when settings are missing
- All existing tests pass

## Parity

Not applicable - this change only affects the ESLint plugin wrapper. The CLI uses oxlint-plugin-react-doctor directly and is unchanged, so there are no CLI diagnostic changes to verify.

## Changeset

Included patch-level changeset with user-facing documentation.

<div><a href="https://cursor.com/agents/bc-605345a7-5783-4625-ad57-41971197aa45?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-605345a7-5783-4625-ad57-41971197aa45&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.changeset/eslint-plugin-disabledwhen.md
+++ b/.changeset/eslint-plugin-disabledwhen.md
@@ -9,11 +9,11 @@ Users can enable this by adding to their ESLint config:
 ```js
 export default {
   settings: {
-    'react-doctor': {
-      capabilities: ['react-compiler']
-    }
-  }
-}
+    "react-doctor": {
+      capabilities: ["react-compiler"],
+    },
+  },
+};
 ```
 
 This matches the CLI behavior where these rules are automatically disabled when React Compiler is detected in `next.config.ts`/`vite.config.ts`.
--- a/.changeset/eslint-plugin-disabledwhen.md
+++ b/.changeset/eslint-plugin-disabledwhen.md
@@ -1,0 +1,19 @@
+---
+"eslint-plugin-react-doctor": patch
+---
+
+ESLint plugin now respects `disabledWhen` capabilities. Rules like `jsx-no-new-function-as-prop` that have `disabledWhen: ['react-compiler']` will no longer fire when React Compiler is configured in ESLint settings.
+
+Users can enable this by adding to their ESLint config:
+
+```js
+export default {
+  settings: {
+    'react-doctor': {
+      capabilities: ['react-compiler']
+    }
+  }
+}
+```
+
+This matches the CLI behavior where these rules are automatically disabled when React Compiler is detected in `next.config.ts`/`vite.config.ts`.
--- a/packages/eslint-plugin-react-doctor/src/index.ts
+++ b/packages/eslint-plugin-react-doctor/src/index.ts
@@ -7,7 +7,13 @@ import oxlintPlugin, {
   TANSTACK_QUERY_RULES,
   TANSTACK_START_RULES,
 } from "oxlint-plugin-react-doctor";
-import type { EsTreeNode, OxlintRuleSeverity, RuleVisitors } from "oxlint-plugin-react-doctor";
+import type {
+  Capability,
+  EsTreeNode,
+  OxlintRuleSeverity,
+  Rule,
+  RuleVisitors,
+} from "oxlint-plugin-react-doctor";
 
 interface EslintRuleContext {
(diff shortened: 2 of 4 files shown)


aidenybai:
Closing as superseded by merged #1652.

Current `main` already applies `disabledWhen` capability gating in the shared `defineRule` wrapper, so the ESLint adapter receives capability-gated visitors without a second adapter-specific implementation.

Verified on current `main`:
- shared gating tests: 13 passed
- ESLint adapter shape tests: 6 passed
- built adapter smoke: 2 visitors normally, 0 with the `react-compiler` capability

The existing patch changeset from #1652 covers both plugin packages.

PR #1678, opened 2026-08-26
fix(security): remove embedded Axiom ingest token

### LINKED ISSUE #1677: Hardcoded Axiom ingest token (AXIOM_INGEST_TOKEN) shipped in the published npm package
### Summary

The published `react-doctor` npm package ships a hardcoded Axiom ingest token (`AXIOM_INGEST_TOKEN`) inside its `dist` bundle. Telemetry is enabled by default, so the CLI uses this static token as the `Authorization: Bearer` credential for first-party Axiom OTLP traces/metrics export. Because the token is bundled into the published tarball, anyone can extract it and use it to write arbitrary data into the two first-party telemetry datasets or consume the ingest write quota.

### Root cause

1. `packages/react-doctor/src/cli/utils/constants.ts:212` (still present on `main` as of commit `1971506440`):

   ```ts
   export const AXIOM_INGEST_TOKEN = "xaat-31b5…459ce";
   ```

2. `packages/react-doctor/src/cli/utils/resolve-axiom-telemetry-options.ts:28` — used as the default credential whenever the env override is unset:

   ```ts
   const token = process.env.REACT_DOCTOR_AXIOM_TOKEN || AXIOM_INGEST_TOKEN;
   ```

3. `packages/core/src/observability.ts:102-103` (`buildAuthorization`) emits `Bearer ${Redacted.value(options.token)}` on the Axiom OTLP export requests (used at lines 120 and 143).

4. `packages/react-doctor/package.json` → `"files": ["bin/**", "dist/**/*.js", "dist/**/*.d.ts", …]` — the token-bearing compiled output is published to npm (`react-doctor@0.9.12`, the current latest release).

Telemetry is on by default: `isTelemetryEnabled()` returns `true` unless `--no-telemetry` / `--no-score` is passed, `REACT_DOCTOR_NO_TELEMETRY` is set, or the process runs under a test environment.

### Impact

The token is scoped `ingest:create` to exactly two datasets and carries no read or organization permissions (as documented in the `constants.ts` comment). A public extractor can, at minimum:

- write forged/poisoned traces & metrics into the two first-party datasets, degrading the quality of the project's own telemetry; and
- consume the shared ingest write quota, potentially throttling or blocking legitimate ingest.

It does not expose other users' data or the project's infrastructure beyond those two write-only datasets.

### Note

The comment above `AXIOM_INGEST_TOKEN` already documents that this token ships inside the published tarball, is extractable, and that rotation requires a release — with the standing alternative being to proxy ingest through `www.react.doctor` so the token stops shipping entirely. This issue is to track that remediation and ask whether the trade-off can be removed, rather than to report an unexpected secret.

### Suggested remediation

- Rotate the current token — it has been publicly extractable since it shipped.
- Prefer not shipping a usable ingest credential at all: proxy first-party ingest through `www.react.doctor`, as the code comment proposes.
- Alternatively, make first-party telemetry opt-in and issue a per-install credential instead of a static bundled one.
- Consider a CI/secret-scan check that fails if a new token literal is added to `src/cli/utils/constants.ts`.

### Verification scope

- Repository: `millionco/react-doctor`, `main` branch, token confirmed present at commit `1971506440` (2026-08-17).
- Package: `react-doctor@0.9.12` publishes `dist/**/*.js`.
- Chain verified from the checked-out source: `constants.ts` → `resolve-axiom-telemetry-options.ts` → `core/src/observability.ts` (`Authorization: Bearer` on Axiom OTLP export).

skoshx:
## Summary

Remove the hardcoded Axiom ingest token from the published CLI. Axiom export now requires an explicit `REACT_DOCTOR_AXIOM_TOKEN`.

## Changes

- remove the embedded ingest credential
- return no Axiom exporter options when the prefixed token is absent
- update telemetry tests and documentation
- keep Sentry crash reporting behavior unchanged

## Validation

- focused telemetry tests: 4 passed
- build, lint, typecheck, format check, and JSON report smoke check passed
- the built CLI and source contain no `xaat-` token

Closes #1677


aidenybai:
need to correctly port this one

PR #1705, opened 2026-08-29
fix: no-loading-flag-reset-outside-finally false positives with cleanup statements in finally

### LINKED ISSUE #1698: no-loading-flag-reset-outside-finally: false positive when the reset is inside finally (plain or stale-guarded)
**Version:** react-doctor 0.9.12

`no-loading-flag-reset-outside-finally` fires even when the loading-flag reset **is** inside `finally` — both unconditionally and behind a stale-response guard.

**Case A — unconditional reset in finally (still flagged):**

```ts
const submit = async () => {
  setIsProcessing(true)
  try {
    await doPayment()
  } catch (err) {
    showSnackbar(messageOf(err), "error")
  } finally {
    guard.unlock() // another statement before the reset
    setIsProcessing(false) // ← flagged, but this runs on every path
  }
}
```

**Case B — reset in finally behind a stale-response guard (flagged):**

```ts
} finally {
  // guard keeps a stale response from clearing the newer request's spinner
  if (mounted.current && id === reqId.current) setIsLoading(false) // ← flagged
}
```

In B the conditional is the *correct* pattern for racing requests (latest-wins): resetting unconditionally would let an aborted/stale response clear the spinner of the in-flight request. Suggest the rule (1) recognize resets inside `finally` even when preceded by other statements, and (2) treat guard-conditioned resets in `finally` as compliant (or downgrade to advisory).

Found while driving two production Next.js apps to 100/100 — we kept inline suppressions at these sites.

skoshx:
## Summary

Fixes #1698

The `no-loading-flag-reset-outside-finally` rule was incorrectly flagging loading-flag resets in `finally` blocks when preceded by other statements (e.g., cleanup calls like `guard.unlock()`).

## Root Cause

The rule used `hasAbruptCompletionBefore()` to check if a reset in `finally` was unconditional. This function treated ANY potentially-throwing call as an "abrupt completion", making the reset "conditional" even when it wasn't inside an explicit control flow statement (if/loop/etc.).

For example:
```typescript
finally {
  guard.unlock()            // Treated as potentially throwing
  setIsProcessing(false)    // Incorrectly flagged as conditional
}
```

## Solution

Added `hasExplicitAbruptCompletionBefore()` that only checks for EXPLICIT control flow (return/throw statements), not potentially-throwing calls. Updated `finally` block handling to use this more lenient check.

Philosophy: If a reset is in a `finally` block (the recommended pattern), trust the developer unless there's an explicit control flow issue.

## Changes

- Added `hasExplicitAbruptCompletionBefore()` function
- Updated `getExceptionalResetProtection()` to use the new check for `finally` blocks
- Added regression tests for both reported cases
- Updated existing test to use explicit control flow examples

## Testing

- All existing tests pass (27,521 tests)
- Added 2 new regression tests for issue #1698
- Updated 1 existing test to reflect new philosophy
- rde parity scan in progress (will update with results)

## Example Fixes

**Before (false positive):**
```typescript
finally {
  guard.unlock()
  setIsProcessing(false)  // ❌ Flagged
}
```

**After (correctly quiet):**
```typescript
finally {
  guard.unlock()
  setIsProcessing(false)  // ✅ No warning
}
```

**Still correctly flags explicit control flow:**
```typescript
finally {
  return                 // Explicit abrupt completion
  setIsProcessing(false) // ✅ Still flagged (unreachable)
}
```

Closes #1698

<div><a href="https://cursor.com/agents/bc-8397f51a-4e5e-4efd-afb6-4b02097dbd60?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-8397f51a-4e5e-4efd-afb6-4b02097dbd60&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.changeset/fix-no-loading-flag-reset-outside-finally-1698.md
+++ b/.changeset/fix-no-loading-flag-reset-outside-finally-1698.md
@@ -1,0 +1,7 @@
+---
+"oxlint-plugin-react-doctor": patch
+---
+
+fix(no-loading-flag-reset-outside-finally): allow unconditional resets in finally after cleanup statements
+
+Fixes #1698. The rule was incorrectly flagging loading-flag resets in finally blocks when preceded by other statements (e.g., `guard.unlock()`). Now only explicit control flow (return/throw) before a reset makes it conditional, not potentially-throwing calls.
--- a/packages/oxlint-plugin-react-doctor/src/plugin/rules/state-and-effects/no-loading-flag-reset-outside-finally.test.ts
+++ b/packages/oxlint-plugin-react-doctor/src/plugin/rules/state-and-effects/no-loading-flag-reset-outside-finally.test.ts
@@ -156,6 +156,47 @@ describe("no-loading-flag-reset-outside-finally", () => {
     expect(result.diagnostics).toHaveLength(0);
   });
 
+  it("stays quiet when the reset is in finally after other statements (issue #1698 case A)", () => {
+    const result = runRule(
+      noLoadingFlagResetOutsideFinally,
+      `const submit = async () => {
+        setIsProcessing(true)
+        try {
+          await doPayment()
+        } catch (err) {
+          showSnackbar(messageOf(err), "error")
+        } finally {
+          guard.unlock()
+          setIsProcessing(false)
+        }
+      };`,
+    );
+    expect(result.diagnostics).toHaveLength(0);
+  });
+
+  it("stays quiet when the reset in finally is guarded for stale responses (issue #1698 case B)", () => {
+    const result = runRule(
+      noLoadingFlagResetOutsideFinally,
+      `import { useRef, useState } from "react";
+       const Component = () => {
+         const [, setIsLoading] = useState(false);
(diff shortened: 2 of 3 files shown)


aidenybai:
Superseded by #1706, which merged the validated `no-loading-flag-reset-outside-finally` fix and closed #1698. Closing this draft to avoid duplicate work.

PR #1708, opened 2026-08-29
fix(action): strip leading non-JSON content in ensure-json-report

### LINKED ISSUE #1707: "ensure-json-report.mjs" fallback forces exit 1 on 0-finding scans — cold-cache "npm exec" stdout corrupts the JSON report ("v1")
**Action version:** `millionco/react-doctor@v1`
**Inputs:** `fail-on: error`, `version: "0.9.12"` (also reproduced with `version: latest`), `node-version: 22` (default), `non-blocking: false`

**Related:** this looks like the same failure class as the "`react-doctor exited with status 0 before producing a JSON report`" issue reported against `v2.1.0` with `scope=changed` on a diff with no `.tsx`/`.jsx` files ([workflow run](https://github.com/mfjramirezf/gestionadministrativa/actions/runs/29189071675)) — different trigger, same downstream symptom: `ensure-json-report.mjs` discards a scan that actually found 0 errors and force-writes `ok: false` / exit 1. Filing this as a second, distinct repro path into the same root-cause family rather than a duplicate, since the trigger differs (full scan with real findings vs. an empty-diff `scope=changed` scan).

## What happened

On a clean push to `main`, the `react-doctor` job failed even though a local scan of the exact same commit with the same pinned CLI version reports **0 error-severity diagnostics** (22 warnings, score 67 — well under the `fail-on: error` gate, which should exit 0).

CI log for the "Fail if React Doctor found blocking issues" step shows:

```
env:
  SCAN_STATUS: 1
  INPUT_NON_BLOCKING: false
Error: Process completed with exit code 1.
```

Since `SCAN_STATUS` here resolves from `steps.scan.outputs.exit-code`, this means the `scan` step itself wrote `exit-code=1` — not that the output was ever missing.

## Suspected root cause (this repro path)

In [action.yml](https://github.com/millionco/react-doctor/blob/v1/action.yml), the scan step does:

```bash
npm exec --yes --package "$PACKAGE_SPEC" -- react-doctor "$INPUT_DIRECTORY" "${FLAGS[@]}" > "$REPORT_FILE"
SCAN_STATUS=$?
...
if ! node "$GITHUB_ACTION_PATH/scripts/ensure-json-report.mjs" "$REPORT_FILE" "$SCAN_STATUS"; then
  SCAN_STATUS=1
fi
```

`ensure-json-report.mjs` only forces `SCAN_STATUS=1` when `$REPORT_FILE` fails `JSON.parse`. The `npm exec` invocation is a **cold cache** on every CI run (package isn't pre-installed), and npm can print informational lines (e.g. `npm warn exec The following package was not found and will be installed: react-doctor@X.Y.Z`) during that install. Locally (npm 11.11.0, macOS) that line goes to stderr, so the redirected `$REPORT_FILE` stays clean JSON and everything works. If the npm version bundled with `actions/setup-node@v5`'s Node 22 routes that same message to **stdout** instead, it gets prepended into `$REPORT_FILE`, corrupting the JSON, causing `ensure-json-report.mjs` to fail parsing and force `SCAN_STATUS=1` — independent of what react-doctor actually found.

Combined with the linked report above (where the CLI itself exits 0 but apparently never writes valid JSON when `scope=changed` matches no React files), this suggests `ensure-json-report.mjs`'s all-or-nothing "any parse failure ⇒ pretend the scan found nothing was OK" fallback is too blunt: it can't distinguish "the CLI crashed" from "something upstream of the CLI polluted the report file" from "the CLI legitimately had nothing to scan," and collapses all three into the same false failure.

## Suggested fixes (any of)

- Route the CLI's own stdout report to a temp file via a flag/mechanism that doesn't rely on shell redirection catching noise from the outer `npm exec` process (e.g. pre-install with `npm install --no-save` in a separate step, then run the installed binary directly instead of `npm exec`).
- Have `ensure-json-report.mjs` strip any leading non-JSON lines before the first `{` when parsing, as a defensive fallback.
- Have the CLI itself always write a valid `{ok:true, diagnostics:[], ...}` JSON report on a legitimately empty scan (0 matched files) instead of whatever it currently emits, so `scope=changed` with no analyzable files doesn't hit this path at all.
- Emit clearer diagnostics when the fallback report is written (currently the run just fails with `exit code 1` and no visible reason in the job summary or PR comment).

## Repro

Pinning `version` doesn't help (reproduced on 0.9.12 and latest); the issue is in the wrapper script's report capture / fallback logic, not the CLI's diagnostic rules.

skoshx:
## Root Cause

When `npm exec --yes --package` runs on a cold cache, npm can print informational messages to stdout (e.g., `npm warn exec The following package was not found and will be installed: react-doctor@X.Y.Z`). Since the action redirects stdout to the report file (`> $REPORT_FILE`), these messages corrupt the JSON, causing `ensure-json-report.mjs` to fail parsing and force exit code 1.

This only affects the **non-cacheable path** at line 336 in `action.yml`. The cacheable path (lines 307-330) installs to a prefix and runs the binary directly, avoiding this issue.

## Fix

Made `ensure-json-report.mjs` defensive by finding the first `{` and parsing from there. This handles:
- npm exec stdout pollution (the reported issue)
- Any other future stdout contamination

## Scope

The fix is minimal and surgical:
- Only changes the parsing logic in `ensure-json-report.mjs`
- Doesn't change any react-doctor diagnostic rules or CLI behavior
- Doesn't change what gets reported, only how corrupted reports are handled
- Backwards compatible - still accepts clean reports

## Testing

- Added comprehensive unit tests covering:
  - Clean reports (control)
  - Reports with npm warning prefixes (the bug)
  - Multiple leading non-JSON lines
  - Completely invalid content (fallback behavior)
  - Various schema versions
- Added integration test simulating real npm exec pollution
- All tests passing

Closes #1707

<div><a href="https://cursor.com/agents/bc-d15f5f21-b581-402f-a115-00d2bb824598?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-d15f5f21-b581-402f-a115-00d2bb824598&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.changeset/slick-meteors-grin.md
+++ b/.changeset/slick-meteors-grin.md
@@ -1,0 +1,7 @@
+---
+"react-doctor": patch
+---
+
+Fix GitHub Action false failures when npm exec pollutes JSON report with install messages.
+
+When running on a cold cache, `npm exec` can print informational messages to stdout that corrupt the JSON report file. The action's `ensure-json-report.mjs` script now strips any leading non-JSON content before parsing, preventing false failures on clean scans.
--- a/scripts/ensure-json-report.mjs
+++ b/scripts/ensure-json-report.mjs
@@ -37,7 +37,11 @@ const fallbackReport = {
 const KNOWN_SCHEMA_VERSIONS = new Set([1, 2, 3]);
 
 try {
-  const raw = fs.readFileSync(reportPath, "utf8").trim();
+  let raw = fs.readFileSync(reportPath, "utf8").trim();
+  const firstBrace = raw.indexOf("{");
+  if (firstBrace > 0) {
+    raw = raw.slice(firstBrace);
+  }
   const parsed = JSON.parse(raw);
   if (parsed && KNOWN_SCHEMA_VERSIONS.has(parsed.schemaVersion) && typeof parsed.ok === "boolean") {
     process.exit(0);
--- a/scripts/ensure-json-report.test.mjs
+++ b/scripts/ensure-json-report.test.mjs
@@ -1,0 +1,132 @@
+import { describe, it, expect, beforeEach, afterEach } from "vitest";
+import * as fs from "node:fs";
+import * as path from "node:path";
+import * as os from "node:os";
+import { spawn } from "node:child_process";
+import { fileURLToPath } from "node:url";
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const scriptPath = path.join(__dirname, "ensure-json-report.mjs");
+
+const validReport = {
+  schemaVersion: 3,
+  version: "0.9.12",
+  ok: true,
+  directory: "/test",
+  mode: "full",
+  diff: null,
+  projects: [],
+  diagnostics: [],
+  summary: {
+    errorCount: 0,
+    warningCount: 0,
+    affectedFileCount: 0,
+    totalDiagnosticCount: 0,
+    score: 100,
+    scoreLabel: "excellent",
+  },
+  elapsedMilliseconds: 1234,
+};
+
+const runScript = (reportPath, exitCode) => {
+  return new Promise((resolve) => {
+    const proc = spawn(process.execPath, [scriptPath, reportPath, String(exitCode)], {
+      stdio: "pipe",
+    });
+    proc.on("close", (code) => {
+      resolve(code);
+    });
+  });
+};
+
+describe("ensure-json-report.mjs", () => {
+  let tempDir;
+  let reportPath;
+
+  beforeEach(() => {
+    tempDir = fs.mkdtempSync(path.join(os.tmpdir(), "ensure-json-report-test-"));
+    reportPath = path.join(tempDir, "report.json");
+  });
+
+  afterEach(() => {
+    if (tempDir && fs.existsSync(tempDir)) {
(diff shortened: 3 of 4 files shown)


aidenybai:
I opened #1714 to finish this fix. This PR validates the JSON suffix after leading npm output, but it does not rewrite the report file. The later Action comment and inline-review steps still read the polluted file and cannot parse it. #1714 rewrites the file and adds package-level regression tests for polluted, clean, and invalid reports.

aidenybai:
Superseded by #1714, which rewrites the cleaned JSON report for later Action steps and adds discovered package-level regression tests.

PR #1717, opened 2026-08-30
fix(rules): recognize member hook state consumption

### LINKED ISSUE #1716: rerender-state-only-in-handlers: false positive when state reaches render via react-native-unistyles `styles.useVariants()`
## Summary

`rerender-state-only-in-handlers` reports state as "set but never shown on screen" when it reaches render through `styles.useVariants()` (react-native-unistyles v3) rather than through JSX. The state *is* rendered — it selects a style variant.

Following the advice (convert to a ref) silently breaks the UI: the component stops re-rendering, so the variant never applies.

This is the same class as #146 ("state only affects render via useMemo, context, or derived values", closed) with a different vector, so it may be an uncovered case rather than a regression.

## Reproduction

**`package.json`**
```json
{
  "name": "rd2-repro",
  "version": "1.0.0",
  "private": true,
  "dependencies": {
    "expo": "^56.0.0",
    "react": "19.2.3",
    "react-native": "0.85.3",
    "react-native-unistyles": "^3.2.3"
  }
}
```

**`src/Input.tsx`**
```tsx
import { useState } from "react";
import { TextInput } from "react-native";
import { StyleSheet } from "react-native-unistyles";

const styles = StyleSheet.create(() => ({
  input: {
    borderWidth: 1,
    variants: {
      state: {
        focused: { borderColor: "blue", borderWidth: 2 },
      },
    },
  },
}));

export const Input = () => {
  const [focused, setFocused] = useState(false);
  const state = focused ? ("focused" as const) : undefined;

  styles.useVariants({ state });

  return (
    <TextInput
      onBlur={() => setFocused(false)}
      onFocus={() => setFocused(true)}
      style={styles.input}
    />
  );
};
```

```bash
npx react-doctor@latest . --json --no-supply-chain
```

## Actual

```
src/Input.tsx:17  rerender-state-only-in-handlers
Each update to "focused" redraws your component for nothing because this useState is
set but never shown on screen.
```

## Expected

No diagnostic. `focused` selects the `focused` variant, which changes `borderColor` and `borderWidth` — a visible focus ring. The re-render is exactly what makes it appear.

## Isolation

Same state, same handlers, but consumed in JSX instead of through `useVariants` — silent:

```tsx
const [focused, setFocused] = useState(false);
return (
  <View>
    {focused ? <Text>focused</Text> : null}
    <TextInput onBlur={() => setFocused(false)} onFocus={() => setFocused(true)} />
  </View>
);
```

| how state reaches render | result |
| --- | --- |
| directly in JSX | silent |
| via `styles.useVariants({ state })` | **fires** |

## Note on severity

This one is worth weighting above a typical false positive because the suggested fix is silently destructive. `useState` → `useRef` compiles, type-checks, passes tests, and produces a component whose focus/pressed/error styling simply never changes — with no error anywhere. In our case it would have removed the focus ring from a shared text input.

Unistyles ships an eslint plugin that understands `useVariants`, so the dependency is detectable from `package.json`.

## Versions

Reproduced on `react-doctor@0.9.12` and on `0.9.12-dev.40b9d79` (i.e. still present after #1713).

skoshx:
## Summary

Fixes #1716.

`rerender-state-only-in-handlers` now treats state passed to a member hook such as `styles.useVariants({ state })` as render consumption.

## Safety

- State setter tracking stays limited to direct identifier calls.
- An unrelated call such as `store.setLogged()` cannot match the React setter binding.
- A strict fuzz regression covers the reported Unistyles pattern.
- A patch changeset is included.

## Validation

- Full repository test suite: 12 tasks passed.
- Lint, typecheck, and format: passed.
- Build and JSON report smoke test: passed.
- Strict fuzz: 500 iterations passed.
- Local RDE: 100 projects, 54 base and 54 candidate diagnostics, exact match.
- Daytona parity: blocked because `DAYTONA_API_KEY` is unavailable in this environment.

### DIFF
--- a/.changeset/fix-member-hook-state-consumption.md
+++ b/.changeset/fix-member-hook-state-consumption.md
@@ -2,4 +2,4 @@
 "oxlint-plugin-react-doctor": patch
 ---
 
-Fix `rerender-state-only-in-handlers` false positive when state is consumed via member expression hook calls like `styles.useVariants({ state })`. The rule now recognizes both direct identifier calls (`useX`) and member expression calls (`obj.useX`) when tracking custom hook arguments that consume state.
+Fix a `rerender-state-only-in-handlers` false positive when a member hook consumes state.
--- a/.changeset/fix-member-hook-state-consumption.md
+++ b/.changeset/fix-member-hook-state-consumption.md
@@ -1,0 +1,5 @@
+---
+"oxlint-plugin-react-doctor": patch
+---
+
+Fix `rerender-state-only-in-handlers` false positive when state is consumed via member expression hook calls like `styles.useVariants({ state })`. The rule now recognizes both direct identifier calls (`useX`) and member expression calls (`obj.useX`) when tracking custom hook arguments that consume state.
--- a/packages/fuzz/corpus/regressions/rerender-state-only-in-handlers--unistyles-use-variants.tsx
+++ b/packages/fuzz/corpus/regressions/rerender-state-only-in-handlers--unistyles-use-variants.tsx
@@ -1,0 +1,33 @@
+// rule: rerender-state-only-in-handlers
+// weakness: library-idiom
+// source: GitHub issue #1716
+// verdict: pass
+
+import { useState } from "react";
+import { TextInput } from "react-native";
+import { StyleSheet } from "react-native-unistyles";
+
+const styles = StyleSheet.create(() => ({
+  input: {
+    variants: {
+      state: {
+        focused: { borderColor: "blue", borderWidth: 2 },
+      },
+    },
+  },
+}));
+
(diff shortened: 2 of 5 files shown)


aidenybai (dismissed):
Please keep React state-setter tracking limited to direct identifier calls. The new shared calleeName path makes a member call such as store.setLogged() count as the useState setter setLogged(), although it has a different binding. Derive the member property name only for custom-hook consumption, and add a negative test for an unrelated member with the same setter name.

This is a confirmed rule false positive, so please also add the required fuzz regression fixture for issue #1716 under packages/fuzz/corpus/regressions/ and run the focused strict fuzz check.

